PR #30, opened 2013-09-10
Include and exclude specific files

Ian-Foote:
Allows the developer to specify individual files to include or exclude from the build.


dessant:
The pr seems to be abandoned and conflicts with master, i have opened an issue for exclude_patterns: https://github.com/kivy/buildozer/issues/245.

PR #185, opened 2015-03-07
allow custom source folders in buildozer.spec

kived:
Allows you to specify custom folders to use when building recipes:

    requirements.source.kivy = ../../kivy

This will set `P4A_kivy_DIR` for the Android target and `kivy_DIR` for the iOS target. The iOS portion is completely untested, as I don't have a Mac or an iOS device.

### DIFF
--- a/buildozer/targets/android.py
+++ b/buildozer/targets/android.py
@@ -464,6 +464,17 @@ class TargetAndroid(Target):
         if not exists(dist_dir):
             need_compile = 1
 
+        source_dirs = {'P4A_{}_DIR'.format(name[20:]):
+                           realpath(expanduser(value))
+                       for name, value in self.buildozer.config.items('app')
+                       if name.startswith('requirements.source.')}
+        if source_dirs:
+            need_compile = 1
+            self.buildozer.environ.update(source_dirs)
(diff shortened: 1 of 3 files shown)


tito:
Can you rebase with the latest master? That would be a good addition :)
(the ios was still based on the previous toolchain, not working anymore)

kived:
Rebase complete :)

buildozer/targets/ios.py
```diff
@@ -128,6 +128,17 @@ def compile_platform(self):
         if last_requirements != ios_requirements:
             need_compile = 1
 
+        source_dirs = {'{}_DIR'.format(name[20:]):
```
tito: the ios toolchain require the name to be upper()

buildozer/targets/android.py
```diff
@@ -464,6 +464,17 @@ def compile_platform(self):
         if not exists(dist_dir):
             need_compile = 1
 
+        source_dirs = {'P4A_{}_DIR'.format(name[20:]):
```
tito: Where the 20 came from?

kived:
Alright, I added `.upper()` for iOS and added comments to explain the `[20:]` part.

PR #192, opened 2015-04-08
Windows fix

jaynakus:
Root detection os.geteuid() does not work on win32 environment and
throws AttributeError. Fixed


tito:
I wonder what you can actually do under windows. Have you been able to use it for anything?

jaynakus:
At least managed to run provided examples :)

PR #197, opened 2015-05-23
check for complete dist instead of dist dir

kived:
This checks for the last file copied into the dist in p4a. It would be best if p4a actually created a specific file to tell that the dist is complete, but that could break compatibility. It might be best to make p4a do that when Buildozer is updated to support @inclement's Python build script.


tito:
Thanks !

PR #206, opened 2015-06-14
make _read_version_subdir return parse('0') instead of [0], otherwise…

denys-duchier:
… it cannot be meaningfully compared to an actual version

I was unable to build apks because build-tools were not being installed.  this patch seems to fix the problem.


tito:
Not tested, i hope you're right :) Thanks you

PR #213, opened 2015-07-04
Fixed bug where an Android package ending with -preview caused an error.

kronenpj:
Today's Android package manifests caused a ValueError exception where one of the packages ended with -preview and int() was called on it.
This will cause buildozer to ignore that package.


DrMoriarty:
Thanks!
I have the same problem

dessant:
This was a package bug which has been fixed, we should probably not pass silently.
https://github.com/kivy/buildozer/issues/217#issuecomment-145337671

PR #250, opened 2015-11-06
fix android.branch option

tshirtman:


dessant:
@tshirtman, could you rebase and merge?

PR #256, opened 2015-12-05
update installation info

kiok46:
Updated the installation info. 
Adding "Android on Ubuntu 15.10  (64bit)" : Tested in my laptop.
Adding `zlib1g:i386` as this package includes debugging symbols for use when debugging applications which use zlib.
I hope everything is correct.

### DIFF
--- a/docs/source/installation.rst
+++ b/docs/source/installation.rst
@@ -17,6 +17,17 @@ If you target Android, you must install at least Cython, few build libs, and a
 Java SDK. Some binaries of the Android SDK are still in 32 bits, so you need
 few 32bits libraries available::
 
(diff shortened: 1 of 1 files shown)


kived:
Did you test all the changes you made to the other platforms? Really, I think we should remove most of those sections anyway - it doesn't make sense to have installation instructions for random Ubuntu versions. Maybe just have 12.04, 14.04, 15.10. 16.04.

tito:
We can at least merge and continue to fix doc, rather than waiting more

kiok46:
@kived  sorry for late reply, I think I missed the notification for the same.
It should be good for 15 and 14 as I have worked on both and needed those extra packages but have not tested for rest.

Issue #245, opened 2015-11-03
Implement source.include_patterns


dessant:
ref https://github.com/kivy/buildozer/pull/30.

udiboy1209:
How difficult would it be for a beginner to buildozer to implement this feature? I have contributed to kivy and also have good enough python knowledge, but I haven't contributed to buildozer so I'm not familiar with the source. But this seems like an interesting feature to code, so I am asking.

inclement:
I think it should be quite doable - there's probably just one place in the code where you need to check and use such a config token, and it will be in the same place as all the similar ones.

udiboy1209:
What should be the priority of the include/exclude options. I mean if files satisfy both include_pattern and exclude_pattern then? And what of files which do not satisfy include_pattern but are in the source dir?

kived:
This has been added in #269

PR #269, opened 2016-01-07
Implement source.include_patterns

udiboy1209:
As per #245 and #30 

Matches files and folders with a list of include patterns similar to exlude_patterns functioning

Priority of include_patterns is above exclude_patterns/exclude_dirs, and below include_exts/exclude_exts

I haven't tested this because I am facing some problems setting up buildozer.


kived:
Good work 👍

PR #274, opened 2016-01-24
put bin/ in builddir if specified in buildozer.spec

jabbalaci:
I prefer to keep generated (thus reproducable) things outside of my project folder. Thus, I added this to my `buildozer.spec` file (example):

```
[buildozer]
builddir = /home/jabba/.kivy_builddirs/kivy_project
```

Now, if I launch `buildozer android debug deploy run`, the local folder `.buildozer` is created in this specified builddir instead of the project's folder. However, `bin/` is still created in the project's folder, littering it with a big `.apk` file.

My patch does the following: if builddir is specified (as seen above), then `bin/` (similarly to `.buildozer/`) will be created in the builddir. This way the project folder has the source files and things necessary for the build process are nicely separated.

Rationale: I keep my projects in my Dropbox folder and I like to keep their sizes to the minimum. Build stuff should go somewhere else.

### DIFF
--- a/buildozer/__init__.py
+++ b/buildozer/__init__.py
@@ -130,6 +130,7 @@ class Buildozer(object):
         except:
             pass
         self.builddir = self.config.getdefault('buildozer', 'builddir', None)
+        self.bin_apk_dir = self.config.getdefault('buildozer', 'bin_dir', None)
 
         self.targetname = None
         self.target = None
@@ -417,17 +418,18 @@ class Buildozer(object):
         self.mkdir(self.global_buildozer_dir)
         self.mkdir(self.global_cache_dir)
 
-        # create local dir
-        specdir = dirname(self.specfilename)
-        if self.builddir:
-            specdir = self.builddir
+        # create local .buildozer/ dir
+        self.mkdir(self.buildozer_dir)
+        # create local bin/ dir
+        self.mkdir(self.bin_dir)
 
-        self.mkdir(join(specdir, '.buildozer'))
-        self.mkdir(join(specdir, 'bin'))
         self.mkdir(self.applibs_dir)
         self.state = JsonStore(join(self.buildozer_dir, 'state.db'))
 
         if self.targetname:
+            specdir = dirname(self.specfilename)
+            if self.builddir:
+                specdir = self.builddir
             target = self.targetname
             self.mkdir(join(self.global_platform_dir, target, 'platform'))
             self.mkdir(join(specdir, '.buildozer', target, 'platform'))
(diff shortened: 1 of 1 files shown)


tito:
Hi, thanks for your patch!
However, i think that must be a separated option, such as bin_dir or something like that. It's not build, and deleting your builddir would also delete all the generate version of your application.

I think bin_dir option would be better either for provisioning another directory from the generated apk.

Would you mind to reflect thoses changes on your PR? That would be cool :D

jabbalaci:
OK, I can do that. But what should it be called? `builddir` has no underscore. Should it be `bindir` or `bin_dir`? It'd work this way:
```
[buildozer]
bin_dir = /home/jabba/.kivy_builddirs/kivy_project
```
Then `/home/jabba/.kivy_builddirs/kivy_project/bin` would be created and the `.apk` files will go there. Is it OK like that?

tito:
bin_dir with underscore. builddir without underscore is a mistake. And no, you don't append bin/ to the bid_dir, that's the point of it :)

jabbalaci:
The pull request is updated. Here is how to use it in the spec file (example):
```
[buildozer]
builddir = /home/jabba/.kivy_builddirs/kivy_project
bin_dir = /home/jabba/.kivy_builddirs/kivy_project
```
Now the `bin/` dir. is created in this folder. If it's not specified, then it's created in the project folder.

tito:
I guess we misunderstood, you don't append bin/ to the bin_dir. The end result must be:

    [buildozer]
    builddir = /home/jabba/.kivy_builddirs/kivy_project
    bin_dir = /home/jabba/.kivy_builddirs/kivy_project/bin

You can even imagine something like:

    bin_dir = /home/jabba/releases/my_projects/

The bin/ suffix is the default of bin_dir, or if bin_dir is specificed, use it as it is

jabbalaci:
OK, corrected. I misunderstood when you wrote "you don't append bin/ to the bin_dir", I thought you meant the spec file.

However, this behaviour is not consistent. In the case of `builddir`, something is appended automagically. In the case of `bin_dir`, nothing is appended. It can be confusing for the users.

jabbalaci:
Should I do something more?

SandervanNoort:
Maybe correct the mistake of the name "builddir", by creating a new config option "build_dir", which will not contain the subfolder "~/.builddir". That way the config "builddir" could be kept for backwards compatibility.

Related, when specifying a builddir, there is currently an error if there is no directory "bin/" in the main kivy source dir.

File "/usr/lib/python2.7/dist-packages/buildozer/targets/android.py", line 749, in build_package
    copyfile(join(dist_dir, 'bin', apk), join(self.buildozer.bin_dir, apk))
  File "/usr/lib/python2.7/shutil.py", line 83, in copyfile
    with open(dst, 'wb') as fdst:
IOError: [Errno 2] No such file or directory

kived:
This looks good to me, if nobody (@tito? :P) has any objections I'll merge this tomorrow for our release.

PR #276, opened 2016-01-26
Corrected Java path on cygwin

brianpanneton:
Allows buildozer to find cygwin java path

### DIFF
--- a/buildozer/targets/android.py
+++ b/buildozer/targets/android.py
@@ -193,7 +193,10 @@ class TargetAndroid(Target):
     def _set_win32_java_home(self):
         if 'JAVA_HOME' in self.buildozer.environ:
             return
-        import _winreg
+        if platform in ('win32'):
+            import _winreg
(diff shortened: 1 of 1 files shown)


buildozer/targets/android.py
```diff
@@ -193,7 +193,10 @@ def _get_available_permissions(self):
     def _set_win32_java_home(self):
         if 'JAVA_HOME' in self.buildozer.environ:
             return
-        import _winreg
+        if platform in ('win32'):
```
kived: This won't work, because `('win32')` isn't a tuple. There's no reason to not just use == here anyway.

buildozer/targets/android.py
```diff
@@ -193,7 +193,10 @@ def _get_available_permissions(self):
...
+            import _winreg
```
dessant: `_winreg` has been renamed to `winreg` in python 3.

tito:
Not fixed since more than one year, closing.

PR #280, opened 2016-02-03
install requires appdirs

cbenhagen:


dessant:
It seems to be a dependency of p4a, along with other things: https://github.com/kivy/python-for-android/blob/98d8c15852876ab7c5d5b0196fdf3b14e2abf015/setup.py#L50.

cbenhagen:
This is true. But buildozer clones p4a and thus will not instal its dependencies. Maybe this should be changed then?

dessant:
Yes, i think there is a plan to publish p4a on pypi, and we could install from there.
For now, maybe we could start intalling with pip from github, rather than cloning.

PR #286, opened 2016-02-11
Ugly hack to fix wrong data being used for target Android API in new …

mixedCase:
…toolchain

The old toolchain passes only the SDK version, which is *not* the same as the target API. This attempts to correct the problem, but should probably be correctly dealt with once the new toolchain becomes the default one.

### DIFF
--- a/buildozer/targets/android_new.py
+++ b/buildozer/targets/android_new.py
@@ -59,7 +59,9 @@ class TargetAndroidNew(TargetAndroid):
                 continue
             elif option == "--sdk":
                 cmd.append("--android_api")
-                cmd.extend(values)
+                values = list(values)
+                values[0] = self.buildozer.config.getdefault('app', 'android.api', self.android_api)
+                cmd.extend(tuple(values))
             else:
                 cmd.extend(args)
 


kived:
The proper fix would be in the android target. There's no need for an ugly hack, the android target does it wrong as well. It passes the sdk value as api and the minsdk value (which doesn't exist) as minapi. The issue is that the old toolchain uses `--sdk` and `--minsdk` to set the API levels.

mixedCase:
Is there any issue open about this? If this is going to be properly fixed soon then I guess this PR is better off closed.

kived:
No, I don't think there is any issue open. The confusion stems from the [`<uses-sdk />` tag in AndroidManifest.xml](http://developer.android.com/guide/topics/manifest/uses-sdk-element.html), which specifies the API levels for your application.

So in p4a old-toolchain, `--sdk` specifies the target API for your application, and `--minsdk` sets the minimum API. But in buildozer.spec, `android.api` and `android.minapi` set those values, while `android.sdk` is the actual SDK version and `android.minsdk` doesn't exist.

kived:
Fixed in #334, but thanks for pointing out the issue. :)

PR #288, opened 2016-02-20
Fix issues with android.p4a_dir spec file property

pastcompute:
- `android.py` uses non-existing `critical()` method when `android.p4a_dir` doesn't exist
- interpret tilde (~) for $HOME in `android.p4a_dir`, also makes consistent with `android.ndk_path`, etc.


akshayaurora:
@pastcompute thanks!

Issue #267, opened 2016-01-07
buildozer should be usable without internet


dessant:
Ported from https://github.com/kivy/kivy/issues/3895.

At the moment, buildozer updates metadata for its android distribution every time it's run. This is fine as long as it's run on a machine with internet, but my build machine isn't supposed to connect to the internet.

Please add a switch to disable the update behavior.
##

Julian-O:
This is being closed, but not forgotten. It has been consolidated into issue #1727.

PR #290, opened 2016-02-20
Add spec option to skip automated update of installed android package

pastcompute:
This patch adds an option to the spec file that lets you disable automated updates of the android sdk for a given project.

Now you can add `android.skip_updates = True` and the build command will skip the call that decides to download updates to Android SDK components.

I was in the middle of a project when the `buildozer android debug deploy` command decided it wanted to download an update to Android SDK 24 even though I didn't need it, and I didn't want to wait an hour for it to come down. So I added this option to fix that.

Note, I think this is different to issue #267 which wants to disable all possible Internet checks; my patch may fix that but at this point I don't know what other tasks may be referencing the Internet...

### DIFF
--- a/buildozer/default.spec
+++ b/buildozer/default.spec
@@ -101,6 +101,11 @@ fullscreen = 1
 # (list) python-for-android whitelist
 #android.p4a_whitelist =
 
+# (bool) If True, then skip trying to update the Android sdk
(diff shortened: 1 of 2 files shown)


inclement:
Looks nice, thanks.

PR #291, opened 2016-02-23
use ios-deploy version 1.7.0

cbenhagen:
buildozer fails with
````
Command failed: make ios-deploy
````
This happens because ios-deploy changed their project structure after version 1.7.0. This issue shows again that we should always use releases of other projects instead of the master branch.


inclement:
Thanks!

PR #316, opened 2016-04-13
trying to fix Kivy install for OS X builds

derPinguin:
I had trouble using buildozer to package my application on my mac book. As it turned out there where some minor bugs like deprecated URLs and commands not working. I also tried to make it a bit more dynamic, so it works with Python 2.7 and Python 3.X and chooses the corresponding Kivy version to download depending on what Python version you are using.

This commit should fix most of the obvious problems. (I tested it for OS X 10.11.3)

### DIFF
--- a/buildozer/targets/osx.py
+++ b/buildozer/targets/osx.py
@@ -47,53 +47,47 @@ class TargetOSX(Target):
         check_call(('unzip', 'master.zip'), cwd=platdir)
         check_call(('rm', 'master.zip'), cwd=platdir)
 
-    def download_kivy(self, cwd, kivy='Kivy2'):
-        self.buildozer.info('Downloading kivy...')
-        if not exists(cwd+'/'+kivy+'.7z'):
-            check_call(
-                ('curl', '-O', '-L',
-                'http://kivy.org/downloads/tests/{}.7z'.format(kivy)),
-                cwd=cwd)
-        if not exists(cwd+'/Keka.app'):
-            if exists('/Applications/Keka.app'):
-                check_call(
-                    ('cp', '-a', '/Applications/Keka.app', './Keka.app'),
-                     cwd=cwd)
-            else:
-                check_call(('curl', '-o',
-                    'http://www.kekaosx.com/release/Keka-1.0.4-intel.dmg'),
-                    cwd=cwd)
-                check_call(
-                    ('hdiutil', 'attach', 'Keka-1.0.4-intel.dmg'),
-                    cwd=cwd)
-                check_call(
-                    ('cp', '-a','/Volumes/Keka/Keka.app',
-                    './Keka.app'), cwd=cwd)
-                check_call(('hdutil', 'detach', '/Volumes/Keka'))
-        check_call(
-            ('Keka.app/Contents/Resources/keka7z',
-            'x', '{}.7z'.format(kivy)), cwd=cwd)
-        check_call(('rm', '-rf', '{}.7z'.format('Kivy2')), cwd=cwd)
-        check_call(('mv', '{}.app'.format(kivy), 'Kivy.app'),cwd=cwd)
+    def download_kivy(self, cwd, py_branch=2):
+        current_kivy_vers = '1.9.1'
+
+        if exists('/Applications/Kivy{}.app'.format(py_branch)):
+            self.buildozer.info('Kivy found in Applications dir...')
+            check_call(('cp', '-a', '/Applications/Kivy{}.app'.format(py_branch), 'Kivy.app'), cwd=cwd)
+
+        else:
+
+            if not exists(cwd+'/Kivy{}.7z'.format(py_branch)):
+                self.buildozer.info('Downloading kivy...')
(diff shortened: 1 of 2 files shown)


buildozer/targets/osx.py
```diff
@@ -47,53 +47,47 @@ def ensure_sdk(self):
...
-                check_call(('hdutil', 'detach', '/Volumes/Keka'))
-        check_call(
-            ('Keka.app/Contents/Resources/keka7z',
-            'x', '{}.7z'.format(kivy)), cwd=cwd)
-        check_call(('rm', '-rf', '{}.7z'.format('Kivy2')), cwd=cwd)
-        check_call(('mv', '{}.app'.format(kivy), 'Kivy.app'),cwd=cwd)
+    def download_kivy(self, cwd, py_branch=2):
+        current_kivy_vers = '1.9.1'
+
+        if exists('/Applications/Kivy{}.app'.format(py_branch)):
```
akshayaurora: This check is done outside of download function, no need to duplicate it here.

buildozer/targets/osx.py
```diff
@@ -47,53 +47,47 @@ def ensure_sdk(self):
...
+            self.buildozer.info('Kivy found in Applications dir...')
+            check_call(('cp', '-a', '/Applications/Kivy{}.app'.format(py_branch), 'Kivy.app'), cwd=cwd)
+
+        else:
+
+            if not exists(cwd+'/Kivy{}.7z'.format(py_branch)):
+                self.buildozer.info('Downloading kivy...')
+                check_call(('curl', '-L', '-o', 'Kivy{}.7z'.format(py_branch),
+                    'http://kivy.org/downloads/{}/Kivy-{}-osx-python{}.7z'.format(current_kivy_vers, current_kivy_vers, py_branch)), cwd=cwd)
```
akshayaurora: pep8

buildozer/targets/osx.py
```diff
@@ -47,53 +47,47 @@ def ensure_sdk(self):
...
+
+            if not  exists('/Applications/Keka.app'):
+                self.buildozer.info('Downloading Keka as dependency (to install Kivy)')
```
akshayaurora: You are removing the local check for Keka.app in cur dir, will this not download keka every time?

buildozer/targets/osx.py
```diff
@@ -47,53 +47,47 @@ def ensure_sdk(self):
...
+            check_call(('mv', 'Kivy{}.app'.format(py_branch), 'Kivy.app'),cwd=cwd)
 
     def ensure_kivyapp(self):
         self.buildozer.info('check if Kivy.app exists in local dir')
-        kivy_app_dir = join(
-            self.buildozer.platform_dir,
-            'kivy-sdk-packager-master', 'osx')
+        kivy_app_dir = join(self.buildozer.platform_dir, 'kivy-sdk-packager-master', 'osx')
+
+        py_branch = sys.version_info[0]
```
akshayaurora: This is not the right way to check python version. We don't want to use the python version that ran the packager, we want the python that is packaged with Kivy.app

buildozer/targets/osx.py
derPinguin: No, it uses the installation in '/Applications'. It will only download and install Keka if it is not found there.

buildozer/targets/osx.py
derPinguin: What Python version do you intend to use other than the one running the packager if Kivy isn't installed yet?
If you want users to use the Python version provided by Kivy then this entire download mechanism is useless, or am I missing something?

buildozer/targets/osx.py
derPinguin: Where do you mean? There is a check in the ensure_kivyapp function, but that only checks the local Kivy dir. This checks if Kivy is installed in '/Applicaions'. If so Kivy can be copied from there and doesn't have to be downloaded.

buildozer/targets/osx.py
akshayaurora: This procedure is for packaging a kivy app.

It would be better to check, if kivy.app does not exist, then check bulldozer.spec for a new `osx.python_version` var or to use the version of python in kivy.app if it  exists.

This has the advantage of allowing the user to package a kivy app for python2 or python3 depending on what they want.

buildozer/targets/osx.py
akshayaurora: we are downloading and installing Keka to current dir 
`check_call( ('hdiutil', 'attach', 'Keka-1.0.4-intel.dmg'), cwd=cwd)` not in /Applicaiton so it will keep checking /Applications and keep downloading kaka again.

derPinguin: Ah, I now do see what you mean, I'll fix that.
This line 'installs it to the current dir': `check_call( ('cp', '-a','/Volumes/Keka/Keka.app', './Keka.app'), cwd=cwd)`

buildozer/targets/osx.py
derPinguin: Sounds good, I'll change that...

PR #323, opened 2016-04-21
use p4a revamp --storage-dir option

kived:
Store p4a revamp files in `.buildozer`. Requires kivy/python-for-android#720


akshayaurora:
@kived 👍

Issue #326, opened 2016-04-24
AttributeError: 'NoneType' object has no attribute 'startswith'


Tungsteno74:
```
# Search for Java keytool (keytool)
#  -> found at /usr/lib/jvm/java-7-openjdk-amd64/jre/bin/keytool
# Install platform
Traceback (most recent call last):
  File "/usr/local/bin/buildozer", line 9, in <module>
    load_entry_point('buildozer==0.32dev', 'console_scripts', 'buildozer')()
  File "/usr/local/lib/python2.7/dist-packages/buildozer-0.32dev-py2.7.egg/buildozer/scripts/client.py", line 13, in main
    Buildozer().run_command(sys.argv[1:])
  File "/usr/local/lib/python2.7/dist-packages/buildozer-0.32dev-py2.7.egg/buildozer/__init__.py", line 992, in run_command
    self.target.run_commands(args)
  File "/usr/local/lib/python2.7/dist-packages/buildozer-0.32dev-py2.7.egg/buildozer/target.py", line 90, in run_commands
    func(args)
  File "/usr/local/lib/python2.7/dist-packages/buildozer-0.32dev-py2.7.egg/buildozer/target.py", line 97, in cmd_update
    self.buildozer.prepare_for_build()
  File "/usr/local/lib/python2.7/dist-packages/buildozer-0.32dev-py2.7.egg/buildozer/__init__.py", line 162, in prepare_for_build
    self.target.install_platform()
  File "/usr/local/lib/python2.7/dist-packages/buildozer-0.32dev-py2.7.egg/buildozer/targets/android.py", line 469, in install_platform
    'android.p4a_dir'))
  File "/usr/lib/python2.7/posixpath.py", line 261, in expanduser
    if not path.startswith('~'):
AttributeError: 'NoneType' object has no attribute 'startswith'

```

Any suggest?

Kozubi:
Same here
Also using new_android giving me this error

PR #327, opened 2016-04-25
Bug correcton when 'p4a_dir' is not set in conf

potens1:
This should correct bug #326


kived:
This was fixed with #328.

PR #375, opened 2016-07-07
Always pass required args to p4a in android_new

inclement:
My previous fix had a mistake; these args must always be passed, not just when calling apk.


inclement:
Merging after this was confirmed to work at https://github.com/kivy/python-for-android/issues/821

PR #385, opened 2016-08-02
Add p4a.local_recipes to buildozer.spec to specify a local recipe dir…

cidermole:
…ectory for the build


inclement:
Looks good, thanks.

PR #403, opened 2016-09-25
Adding README.rst entries for how to use buildozer with python3

andyDoucette:
Thanks to rafalo13 and inclement for suggestions on how to get this to work.  I'm just documenting it for others to use.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -59,11 +59,11 @@ The pip package does not yet support python3.
     buildozer init
 
 #. Edit the buildozer.spec file to fit your application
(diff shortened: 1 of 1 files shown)


andyDoucette:
If bug fixes in code are an option, that's even better! :)  I invite you to edit my pull request as desired and take what you like from it.  Sounds like you'll just be removing a couple lines and delaying the update until after the bug is fixed?  Sounds good to me.

Mr-Binary:
Thanks for this, I currently use the official approach which use the hostpython2 to package, my code is written in python3, will it be any problem?

tito:
Thanks you

Issue #392, opened 2016-08-17
Python 3: # Aidl cannot be executed AttributeError: 'module' object has no attribute 'maxint'


hartred:
buildozer fails when running "buildozer android debug".
Used virtualenv with python 3.4 interpreter.
```
# Aidl cannot be executed
Traceback (most recent call last):
  File "/home/kivy/projects/kivy_tutor/bin/buildozer", line 11, in <module>
    sys.exit(main())
  File "/home/kivy/projects/kivy_tutor/lib/python3.4/site-packages/buildozer/scripts/client.py", line 13, in main
    Buildozer().run_command(sys.argv[1:])
  File "/home/kivy/projects/kivy_tutor/lib/python3.4/site-packages/buildozer/__init__.py", line 1020, in run_command
    self.target.run_commands(args)
  File "/home/kivy/projects/kivy_tutor/lib/python3.4/site-packages/buildozer/target.py", line 90, in run_commands
    func(args)
  File "/home/kivy/projects/kivy_tutor/lib/python3.4/site-packages/buildozer/target.py", line 100, in cmd_debug
    self.buildozer.prepare_for_build()
  File "/home/kivy/projects/kivy_tutor/lib/python3.4/site-packages/buildozer/__init__.py", line 174, in prepare_for_build
    self.target.install_platform()
  File "/home/kivy/projects/kivy_tutor/lib/python3.4/site-packages/buildozer/targets/android.py", line 500, in install_platform
    self._install_android_packages()
  File "/home/kivy/projects/kivy_tutor/lib/python3.4/site-packages/buildozer/targets/android.py", line 422, in _install_android_packages
    self._check_aidl(v_build_tools)
  File "/home/kivy/projects/kivy_tutor/lib/python3.4/site-packages/buildozer/targets/android.py", line 450, in _check_aidl
    if sys.maxint > 2**32:
AttributeError: 'module' object has no attribute 'maxint'
```

user5061:
The bug is on [this line](https://github.com/kivy/buildozer/blob/master/buildozer/targets/android.py#L450). According to [Stack Overflow](http://stackoverflow.com/questions/13795758/what-is-sys-maxint-in-python-3) `sys.maxint` was removed in Python3 and there are some solutions. However according to @inclement: 

> That seems like pretty weird code in general, but in any case it could check platform.architecture() instead. I might just remove that if/else entirely. Leave only the else.

user5061:
[...]

PR #409, opened 2016-10-08
Fixed 64 bit detection (it failed under python3)

inclement:


inclement:
Will fix https://github.com/kivy/buildozer/issues/392

imiten:
even after the proposed fix it gives error of you might have missed to install 32bit libs.

inclement:
I think that error is intended, although it might be out of date, as aidl probably used to need to be 32 bit. If this is no longer the case, we can just remove the error as it's more confusing than helpful.

Issue #431, opened 2016-12-04
compiling crash


nesergen:
When I compile APK with buildozer with property 
```
# (list) Android additionnal libraries to copy into libs/armeabi
android.add_libs_armeabi_v7a = libs/android-v7/*.so
```
the ERROR occure

```
# Package the application
# Search and copy libs for armeabi-v7a
Traceback (most recent call last):
  File "/usr/local/bin/buildozer", line 11, in <module>
    sys.exit(main())
  File "/usr/local/lib/python2.7/dist-packages/buildozer/scripts/client.py", line 13, in main
    Buildozer().run_command(sys.argv[1:])
  File "/usr/local/lib/python2.7/dist-packages/buildozer/__init__.py", line 1027, in run_command
    self.target.run_commands(args)
  File "/usr/local/lib/python2.7/dist-packages/buildozer/target.py", line 90, in run_commands
    func(args)
  File "/usr/local/lib/python2.7/dist-packages/buildozer/target.py", line 102, in cmd_debug
    self.buildozer.build()
  File "/usr/local/lib/python2.7/dist-packages/buildozer/__init__.py", line 210, in build
    self.target.build_package()
  File "/usr/local/lib/python2.7/dist-packages/buildozer/targets/android.py", line 637, in build_package
    for fn in self.buildozer.file_matches(patterns):
TypeError: 'NoneType' object is not iterable

```
1) for what purposes this property (Android additionnal libraries to copy into libs/armeabi) needs ?
2) FROM where and TO where must be copied the *.so files ?

inclement:
I've no idea exactly what has gone wrong here, but the file_matches function looks like this:

    def file_matches(self, patterns):
        from glob import glob
        result = []
        for pattern in patterns:
            matches = glob(expanduser(pattern.strip()))
            if not matches:
                return
            result.extend(matches)
        return result

This clearly can return None when it shouldn't, but I'm not sure what the author intended, whether the `return` should be `continue` (seems likely), or whether it should at least `return []`.

inclement:
I made a PR to fix it (linked above). If you need a quick fix, you could try out that change.

nesergen:
I figured out how it works. But to fix I need to know from where the files must be copied

nesergen:
[...]

PR #432, opened 2016-12-04
Fix file_matches to never return None

inclement:
Should fix https://github.com/kivy/buildozer/issues/431 , as the deleted code just makes the function return None when presumably an empty list is fine and it should just continue.

Making a PR rather than a straight change in case I'm missing something about what's intended here.


tito:
Thanks you!

PR #446, opened 2017-01-05
spec file: use portrait orientation by default

rafalo1333:
99% of apps are portrait oriented, only advanced games are using landscape by default, p4a should follow


tito:
Agreed.

PR #447, opened 2017-01-05
spec file: dont use fullscreen by default

rafalo1333:
Most of the Android apps are not fullscreen, statusbar is visible. Default should follow that rule. Only games/navigations are fullscreen by default. This setting makes new users think Kivy is meant only for games, or p4a cant produce app with normal, statusbar visible UI.


tito:
Agreed.

PR #455, opened 2017-01-12
Fix Py3 Incompatable str + bytes issue.

FeralBytes:
Found while using buildozer with py3 on old-toolchain as $ buildozer android debug

### DIFF
--- a/buildozer/__init__.py
+++ b/buildozer/__init__.py
@@ -788,7 +788,10 @@ class Buildozer(object):
                   '"..", "_applibs")] + sys.path\n')
         with open(main_py, 'rb') as fd:
             data = fd.read()
-        data = header + data
+        if IS_PY3:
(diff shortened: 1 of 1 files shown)


buildozer/__init__.py
```diff
@@ -788,7 +788,10 @@ def _add_sitecustomize(self):
                   '"..", "_applibs")] + sys.path\n')
         with open(main_py, 'rb') as fd:
             data = fd.read()
-        data = header + data
+        if IS_PY3:
```
matham: A better solution I suspect is to write 
```py
header = (b'import sys, os; '
                   b'sys.path = [os.path.join(os.getcwd(),'
                   b'"..", "_applibs")] + sys.path\n')
```

FeralBytes:
See here: https://github.com/kivy/buildozer/pull/456

PR #456, opened 2017-01-12
Fix Py3 Incompatable str + bytes issue.

FeralBytes:
Found while using buildozer with py3 on old-toolchain as $ buildozer android debug

### DIFF
--- a/buildozer/__init__.py
+++ b/buildozer/__init__.py
@@ -826,16 +826,11 @@ class Buildozer(object):
             #self.error('Unable to patch main_py to add applibs directory.')
             return
         
-        with open(main_py, 'rb') as fd:
-            data = fd.read()
-        if IS_PY3:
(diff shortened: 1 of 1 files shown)


FeralBytes:
Change was also tested and worked on my system.

FeralBytes:
@matham fixes in accordance with your suggestion.

matham:
Sorry for not being explicit, but you don't need the if statement. It should always have the `b` prefix. The reason it works in py2 without is that it's automatically bytes. But it won't hurt having the `b` prefix and will reduce the complexity as you don't need the if and line duplication.

FeralBytes:
@matham Fixed as noted. Also tested and works.

FeralBytes:
Sorry about the pep8 issue.

matham:
It's ok. I just didn't want to ask for another fix :D

PR #458, opened 2017-01-14
android_new: add "android.arch" config option

ZingBallyhoo:
Allow choosing the arch to build for

Usage for arm64:
```
android.arch = arm64-v8a
```


inclement:
Looks good, thanks!

PR #460, opened 2017-01-16
android_new: add 'symlink-java-src' option

ZingBallyhoo:
Adds ability to set the p4a 'symlink-java-src' option.

Don't know how useful this actually is, but I plan to use it, so I might as well make a PR


tito:
Can't merge this PR, you add a non-configurable behavior. Symlink the java src break the sandboxing (if i don't mistake). If it's an option, then ok.

PR #464, opened 2017-02-03
Add p4a.force-build option

ZingBallyhoo:
To go along with: [https://github.com/kivy/python-for-android/pull/1006](https://github.com/kivy/python-for-android/pull/1006)


tmarplatt:
Among its other purposes, this would be a very useful feature for Recipe implementators running buildozer (like me 🙂).

PR #465, opened 2017-02-04
android_new: change skip_update to skip all updates

ZingBallyhoo:
In current master. android.skip_updates only skips updating platform-tools and tools, and will still update the build-tools. This is a problem for me at least, because Buildozer will auto update my build tools from 19.1.0, and 19.1.0 is the only version I can get to build correctly.

This is just a quick fix so that it will do no updates with the option enabled


tito:
Thanks you!

PR #476, opened 2017-02-27
android / android_new: Work towards Windows builds

ZingBallyhoo:
#### Fixes:

* Buildozer.checkbin looking for files without an extension (without '.exe' so it would never find Git or Cython)
* NDK, SDK and Ant downloading now works.
* Formatting fixes in ```toolchains/android.py```

#### Dependencies:

Ports of the normal Linux P4A dependencies can be downloaded from here: http://gnuwin32.sourceforge.net/packages.html

* GZip: http://gnuwin32.sourceforge.net/packages/gzip.htm
* Tar: http://gnuwin32.sourceforge.net/packages/gtar.htm
* UnZip: http://gnuwin32.sourceforge.net/packages/unzip.htm

#### Issues

Buildozer crashes when trying to do Buildozer.cmd(get_stdout=True), as it is unable to get get_stdout: (below is an error from when it tries to list the SDK)
```
  File "C:\Python35\lib\site-packages\buildozer-0.33.dev0-py3.5.egg\buildozer\targets\android.py", line 355, in _android_list_sdk
    for x in available_packages.splitlines()
AttributeError: 'NoneType' object has no attribute 'splitlines'
```
But, the actual exception is:
```
  File "C:\Python35\lib\site-packages\buildozer-0.33.dev0-py3.5.egg\buildozer\__init__.py", line 307, in cmd
    readx = select.select([fd_stdout, fd_stderr], [], [])[0]
OSError: [WinError 10038] An operation was attempted on something that is not a socket
```

There are definitely more issues that happen later down the line, but I want to get this one sorted out first.

### DIFF
--- a/buildozer/__init__.py
+++ b/buildozer/__init__.py
@@ -243,8 +243,13 @@ class Buildozer(object):
         self.debug('Search for {0}'.format(msg))
         if exists(fn):
             return realpath(fn)
-        for dn in environ['PATH'].split(':'):
-            rfn = realpath(join(dn, fn))
+        is_win = sys.platform in ('win32', 'cygwin')
+        path = environ['PATH'].split(';' if is_win else ':')
+        for dn in path:
+            if is_win:
+                rfn = realpath(join(dn, fn+".exe"))
+            else:
+                rfn = realpath(join(dn, fn))
             if exists(rfn):
                 self.debug(' -> found at {0}'.format(rfn))
                 return rfn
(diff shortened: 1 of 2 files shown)


inclement:
Regarding the p4a parts, matham got p4a apparently working on Windows, see https://github.com/kivy/python-for-android/pull/994.

ZingBallyhoo:
@inclement Meant to say ```Buildozer``` under the issues section, not ```P4A```, in case that it caused any confusion. I'll take a look at the ```P4A``` PR soon.

tito:
I'm concerned about thoses changes. I would like not to merge the style fixes, (lines are too long for examples). You also fixed others style changes (such as adding space after ,) but don't do it on your part of the patch ^^

My point would be: just to a pr concerning thoses changes, not styles changes. (it can also break other current PR). Then after doing the PR merge work, i'll run a yapf on it.

PR #508, opened 2017-06-05
Add 404 status code handling on kivy download

SecretObsession:
Since the Kivy Downloads section is down due to https://github.com/kivy/kivy/issues/5211 the OSX build target has been failing if you are missing the Kivy dependency. (It's been a couple days since it's been taken down, and troubleshooting this is pretty annoying for any devs who clean their build environment.)

This PR will handle the case when the Kivy dependency cannot be downloaded. It will cover the case where the server returns 404, (such is the case now). It will also cover when a user specifies an incorrect kivy version in osx.kivy_version.


tito:
Thanks you!

Issue #519, opened 2017-06-30
The "android" command is deprecated


mimi89999:
When building a Buildozer app with the latest SDK, I get:
```
# Android SDK found at /home/michel/android-sdk
# Android NDK found at /home/michel/android-sdk/ndk-bundle
# Run '/home/michel/android-sdk/tools/android list sdk -u -e'
# Cwd /home/michel/.buildozer/android/platform
*************************************************************************
The "android" command is deprecated.
For manual SDK, AVD, and project management, please use Android Studio.
For command-line tools, use tools/bin/sdkmanager and tools/bin/avdmanager
*************************************************************************
Timed out waiting for input.
To suppress this prompt, run with --use-sdk-wrapper or set USE_SDK_WRAPPER.
# Command failed: /home/michel/android-sdk/tools/android list sdk -u -e
# 
# Buildozer failed to execute the last command
# The error might be hidden in the log above this error
# Please read the full log, and search for it before
# raising an issue with buildozer itself.
# In case of a bug report, please add a full log with log_level = 2
```

Buildozer should use sdkmanager commands now.

misl6:
For a quickly workaround just go back to "SDK Tools, Revision 25.3.0"

mimi89999:
This won't help since I am building on F-Droid...

lzps:
```bash
cd ./.buildozer/android/platform/python-for-android-master/
git pull origin master

export USE_SDK_WRAPPER=True
```

inclement:
This should be fixed for sdl2 by https://github.com/kivy/python-for-android/pull/1071 , but it needs testing.

comodoro:
I updated to current repository version:

`pip install git+https://github.com/kivy/buildozer.git`

and have version `buildozer==0.34.dev0`

but still encountered the prompt

inclement:
@comodoro You might need to clean any exsisting downloaded SDK.

comodoro:
Confirmed, I was using my existing up-to-date SDK (which does not work anyway), but deleting ~/.buildozer has helped (with this particular prompt)

Ryan-ED:
Is downgrading still the only option to fix this issue?

inclement:
Closed by #806

PR #523, opened 2017-07-03
omit package list dowload with android.skip-update

lufebe16:
When specifiying android.skip-update in buildozer.spec current versions download the sdk package list, and then skip the update correctly. Its obvious that the download ist useless.

Indirectly it is also the reason of issue request #519.

### DIFF
--- a/buildozer/targets/android.py
+++ b/buildozer/targets/android.py
@@ -403,33 +403,37 @@ class TargetAndroid(Target):
         if not os.access(self.android_cmd, os.X_OK):
             self.buildozer.cmd('chmod ug+x {}'.format(self.android_cmd))
 
-        # 1. update the tool and platform-tools if needed
-        packages = self._android_list_sdk()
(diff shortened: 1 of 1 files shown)


tito:
Actually, with your changes, i believe the very first time it will be used (for an existing buildozer.spec), it will just not work, as it doesn't try to install the initial version.

Skip updates != skip initial installation. Or is there something i didn't understand?

lufebe16:
Its a long time ago. I didn't even remember the pull request. The facts are as following:

I developed a build for fdroid. But they do not accept a build system, that does downloads of its own. So I decided to download the google packages of my own (to the cache) before running buildozer. And specified 'skip-update'. In that szenario there is no 'first time'.

Afterwards I decided to use p4a directly, as buildozer in fact adds another level of complexity usless to a pure android project. From my point of view the pull request is now obsolete.

Thanks for Your collaboration.

PR #572, opened 2017-10-11
Some fixes in old android target

rnixx:
* Properly decode lines in targets.android.TargetAndroid._update_libraries_references
* Fix path generation for android release apk name


Zen-CODE:
If you take out the .gitignore, this looks good to me. After that, I will test and merge but the .gitignore needs to go :)

Thanks a lot. Appreciated

rnixx:
I removed then changes in .gitignore

rnixx:
Since there are a lot of superfluous commits in this PR you might just cherry-pick https://github.com/kivy/buildozer/pull/572/commits/075e11ab3b61cffc22a977ecee2f7eb826c0a686 and https://github.com/kivy/buildozer/pull/572/commits/a8b0d3e3fc9e46c7d6aa2cf8904f4267392f26fa

Zen-CODE:
I have taken out the changes around reading the content as they use 'unicode', and 'unicode' is not recognized in Python 3, so would not break Py3 support.

rnixx:
imho this is still more elegant than the try/except block

Zen-CODE:
It may well be, but then that can be done in a separate PR. It is not something I could test right now and would need those additional changes accompanyoing it.. But I would like the other changes, which I am testing now and will merge if OK..

rnixx:
@Zen-CODE OK, I'm busy with other stuff as well right now, will open another PR then

Zen-CODE:
Thanks :-)

PR #588, opened 2017-11-28
add p4a.port config option; to allow specifiying webview port

replabrobin:
The buildozer android_new target should allow the webview port to be specified. I tested with flask and it seemed to work.


tito:
Thanks you! (you missed to changes the default.spec too

PR #597, opened 2017-12-05
fixed auto build type (ant or gradle) detection

ibobalo:
fixed auto build type (ant or gradle) apk detection. existence of build.gradle is not reliable method, it may be created by some recipes and is not used by python for android.

### DIFF
--- a/buildozer/targets/android.py
+++ b/buildozer/targets/android.py
@@ -272,7 +272,7 @@ class TargetAndroid(Target):
 
     def _get_android_ndk_url(self):
         import re
-        _version = int(re.search('(.+?)[a-z]', self.android_ndk_version).group(1))
+        _version = int(re.search('([0-9]+)[a-z]?', self.android_ndk_version).group(1))
 
         if platform in ('win32', 'cygwin'):
             # Checking of 32/64 bits at Windows from: http://stackoverflow.com/a/1405971/798575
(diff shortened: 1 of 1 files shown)


inclement:
Thanks, looks good. I'll try to test and merge at the weekend. If anyone else would like to test it and confirm it works, that would be welcome.

tito:
I can't merge this PR as it contains other changes than the one PR targetted.
Also, copying multiple times doesn't fix the detection :)

master got a fix about it

ibobalo:
hi
sorry but your method of build type detection is still not reliable.
file build.gradle may exist even for non graddle builds. for example if p4a
recipe has own build.gradle like sdl2 recipe has. then you will get copy
error tryint to copy apk with graddle name pattern instead and named apk

2017-12-15 16:57 GMT+02:00 Mathieu Virbel <notifications@github.com>:

ibobalo:
hi
sorry but your method of build type detection is still not reliable.
file build.gradle may exist even for non graddle builds. for example if p4a recipe has own build.gradle like sdl2 recipe has. then you will get copy error tryint to copy apk with graddle name pattern instead ant named apk
also i apologize for multiple commits in one pull request, its a problem of github PRs. it just merges two PRs

AndreMiras:
Hi @ibobalo I can tell that after updating to buildozer==0.34 I still get the issue, can you provide another pull request built on top of 0.34 with the reliable method you describe, so I can tell you.

ibobalo:
hi

sorry, its old story. i'm not in kivy/buildozer/python context now and have
no spare time to remember.


ср, 4 лип. 2018 о 16:13 Andre Miras <notifications@github.com> пише:

Issue #602, opened 2017-12-17
FileNotFoundError at end of build when copying APK file.


metamarcdw:
Using buildozer VM and Python 3. Running a simple test app copied straight from https://github.com/kivy/kivy/wiki/Community-Guidelines. APK seems to build and work if copied manually to the phone, but deploy does not work. Copy file fails with FileNotFoundError.

[DEBUG log](https://gist.github.com/metamarcdw/0d8001d9559211b5445ad4af037d46c5)
[buildozer.spec](https://gist.github.com/metamarcdw/8d92c91fd9c48f9721df10964bef98f5)

metamarcdw:
Nevermind. Did a git pull, deleted .buildozer/ and started over. Everything seems to be working now.

metamarcdw:
Reopening this issue. Turns out that this error does still happen, whenever the build is done using Ant as opposed to Gradle.

inclement:
Closing as probably fixed now, ant is not supported and the issue was likely resolved anyway.

PR #609, opened 2017-12-26
Allow to force build type (gradle/ant) in buildozer.spec

maho:
I had situation when I had to force to NOT USE gradle. It was false positive detected, and in effect - build failed.


inclement:
Thanks for the contribution. Would you mind calling the buildozer.spec option something like `android.java_build_tool` to match the p4a option, and placing the setting in the default.spec with the other `android.___` settings?

Also, did you make a p4a issue about this issue? We should be able to improve the autodetection if it's wrong sometimes (it's really just a best guess as it stands).

metamarcdw:
This may also serve as a workaround for my issue #602. Hope to see it merged soon!

mahomahomaho:
@inclement: I don't mind renaming the option. Just pushed the change

Re p4a issue - I don't think it's p4a issue. I rather find it as buildozer issue - wrongly detected tool used in p4a by existence (or not) of some files. 

Also, it's very hard to reproduce the error :). When I clean ~/.buildozer ./.buildozer, then sometimes p4a uses gradle, sometimes p4a uses ant, I don't see deterministic pattern in it.

buildozer/default.spec
```diff
@@ -146,6 +146,9 @@ fullscreen = 0
 # bootstrap)
 #android.gradle_dependencies =
 
+# (str) build type ('auto' - default/auto detect, 'ant' - use ant, 'gradle' - use gradle)
+#android.java_build_tool = 'auto'
```
inclement: Am I right in thinking that `auto` should not actually have the `'` around it here?

inclement:
Looks good - just one question annotated before merging.

Thanks!

mahomahomaho:
could someone merge it to trunk?

tito:
This PR is now outdated, and sdl2 bootstrap handle build type correctly.

PR #612, opened 2017-12-29
Allow adding Java activities to the manifest

bauerj:
This depends on kivy/python-for-android#1213.

This feature will allow users to add their own activities to the manifest file. It can be used to start Java activities using pyjnius. See kivy/python-for-android#1213 for more information.


ecdsa:
this works for me

tito:
Thanks you!

PR #630, opened 2018-02-27
Quote --storage-dir value

bionoid:
Untested

https://i.imgur.com/BfRhQTA.png


inclement:
Looks reasonable, thanks. I'll try to test before merging, since I've been surprised before.

PR #653, opened 2018-04-16
fix prerelease version

marceloneil:
Fixes an issue where prerelease versions were parsed correctly, but not converted to a string correctly.

For example:
Build tools version "28.0.0-rc1" was parsed as "28.0.0rc1", resulting in the following error:
```bash
# Aidl not found, please install it.
```

This might be a poor fix, and could cause issues elsewhere, so I'm happy to make any necessary changes.

### DIFF
--- a/buildozer/libs/version.py
+++ b/buildozer/libs/version.py
@@ -255,7 +255,7 @@ class Version(_BaseVersion):
 
         # Pre-release
         if self._version.pre is not None:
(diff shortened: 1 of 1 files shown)


tito:
I guess they changed the prerelease tag with a minus then. I'm a little bit afraid that it may break if one day they remove it. Anyway for now, let's merge. Thanks you!

Issue #625, opened 2018-02-25
 ERROR: /home/ubuntu/.buildozer/android/platform/apache-ant-1.9.4/bin/ant failed!


MichaelCurrie:
I get this error running the `buildozer --verbose android debug` command.

![image](https://user-images.githubusercontent.com/5267078/36643885-42f6023c-1a84-11e8-9c4f-c753d72afc4e.png)

```
Compiling /home/ubuntu/github/alterx/android_deployment/.buildozer/android/platform/build/dists/AlterX/private/lib/python2.7/xml/sax/xmlreader.py ...
[INFO]:    Selecting java build tool:
[INFO]:    Detected highest available build tools version to be 19.1.0
[INFO]:        Building with ant, as the highest build-tools-version is only 19.1.0
[DEBUG]:   -> running ant debug
[DEBUG]:        Buildfile: /home/ubuntu/github/alterx/android_deployment/.buildozer/android/platform/build/dists/AlterX/build.xml
[DEBUG]:
[DEBUG]:        -set-mode-check:
[DEBUG]:
[DEBUG]:        -set-debug-files:
[DEBUG]:
[DEBUG]:        -check-env:
[DEBUG]:         [checkenv] Android SDK Tools Revision 22.6.2
[DEBUG]:         [checkenv] Installed at /home/ubuntu/.buildozer/android/platform/android-sdk-20
[DEBUG]:
[DEBUG]:        -setup:
[DEBUG]:             [echo] Project Name: AlterX-1.10.1
[DEBUG]:          [gettype] Project Type: Application
[DEBUG]:
[DEBUG]:        -set-debug-mode:
[DEBUG]:
[DEBUG]:        -debug-obfuscation-check:
[DEBUG]:
[DEBUG]:        -pre-build:
[DEBUG]:             [copy] Copying 25 files to /home/ubuntu/github/alterx/android_deployment/.buildozer/android/platform/build/dists/AlterX/tmp-src
[DEBUG]:
[DEBUG]:        -build-setup:
[DEBUG]:        [getbuildtools] Using latest Build Tools: 19.1.0
[DEBUG]:             [echo] Resolving Build Target for AlterX-1.10.1...
[DEBUG]:        [gettarget] Project Target:   Android 4.4.2
[DEBUG]:        [gettarget] API level:        19
[DEBUG]:             [echo] ----------
[DEBUG]:             [echo] Creating output directories if needed...
[...]

Issue #679, opened 2018-06-18
Continuous integration testing


AndreMiras:
Would you guys be interested if I picked that one up?
The idea would be to start simple verifying few things on every pushes and then build on top, e.g.

- tox (linter, multi env...)
- verify install OK on Python2
- verify install OK on Python3

This could be done through Docker (https://github.com/kivy/buildozer/issues/589) for a completely isolated environment or not, up to you.

Ping me if you're interested.

AndreMiras:
Pull request is ready https://github.com/kivy/buildozer/pull/794 :tada:

PR #681, opened 2018-06-18
buildozer Dockerfile, fixes #589

### LINKED ISSUE #589: Feature Request: Dockerfile or image in dockerhub
there's an offical VM, 
I was looking for offical dockerhub image, cause it's much easier to operate (and automated, i.e. travis-ci and such) 

Are some technical diffeculties that I'm missing or just no on had time to invest in it ? what's actully needed to make it "offical" ?
I see some people have something working (I guess, didn't tried it yet)
https://github.com/tshirtman/Buildozer-docker/blob/master/Dockerfile

AndreMiras:
Buildozer Dockerfile, fixes #589

Installs buildozer and dependencies, documents usage.

To build the image you first need to `cd` to where the `Dockerfile` is,
then run:
```sh
docker build --tag=buildozer .
```
Then you can use the container e.g. with:
```sh
docker run --volume "$(pwd)":/home/user/hostcwd buildozer --help
```
The above command mounts host current working directory in container using `--volume` option.
Also added workaround for #625

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,55 @@
+# Dockerfile for providing buildozer
+# Build with:
+# docker build --tag=buildozer .
+# In order to give the container access to your current working directory
+# it must be mounted using the --volume option.
+# Run with (e.g. `buildozer --version`):
+# docker run --volume "$(pwd)":/home/user/hostcwd buildozer --version
+# Or for interactive shell:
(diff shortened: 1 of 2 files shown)


fruch:
will it have a CI to upload (or even test) those images on every new version (or every Pull Request) ?

AndreMiras:
Not for now, and the CI testing via Docker is tracked in a different ticket #679.
But if we realize it breaks too often, then maybe we want to CI that one too.

inclement:
Thanks!

PR #711, opened 2018-09-02
Updates Python 3 install instructions, fixes #710

### LINKED ISSUE #710: Python3Crystax instructions in README
The instructions for python3 support on android don't seem up to date (https://github.com/kivy/buildozer/blob/master/README.rst#installing-buildozer-with-python3-support)
- I'm not sure point 1 is still relevant
- It seems the "android_new"->"android" change in the last bullet point was done without checking if the result still made sense, i think the whole point can be deleted.

AndreMiras:
Refactors Python 2/3 install instructions.
Makes explicit difference between host and target Python.
Both host Python versions are supported seamlessly while target Python
requires different actions depending on the version.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -14,15 +14,18 @@ project, and for iOS via the kivy-ios project. iOS and OSX are still under work.
 
 For Android: please have a look at `Android-SDK-NDK-Informations
 <https://github.com/kivy/kivy/wiki/Android-SDK-NDK-Informations>`_. Please note that
-the default SDK/NDK coded in Buildozer works for Python 2.
+the default SDK/NDK coded in Buildozer works for target Python 2.
+For target Python 3 see `Installing Buildozer with target Python 3 (CrystaX)
+<#installing-buildozer-with-target-python-3-crystax>`_.
+For host Python, both versions are seamlessly supported.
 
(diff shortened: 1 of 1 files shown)


tshirtman:
Looks good to me!

tshirtman:
Just a nit picking, it's better if the "-" lines under titles are the same length as the title itself, as it's what ReStructuredText specify, but i'm not sure it'll make a visible difference in output.

AndreMiras:
Thanks @tshirtman for reviewing. For the `-` I think I kinda get what you mean. You mean space+parenthesis e.g. ` (` should be converted to two dashes i.e. `--` ? Is this what you meant?
I was also not sure at first so I did a live test and fixed it before doing the pull request.
You can try the end result on the README of my fork/branch here https://github.com/AndreMiras/buildozer/tree/feature/ticket710_update_readme_python3_instructions#installing-buildozer-with-target-python-3-crystax
It seems to work for me, give it a try playing with both link/anchor and let me know if you think it's OK.

tshirtman:
Sorry, i should have been clearer, i mean the dashes need to continue up to the end of the line, (so under the final ":").
```
Installing Buildozer with target Python 3 (CrystaX):
----------------------------------------------------
```

PR #740, opened 2018-11-06
workaround for working in WSL

tshirtman:
this makes it easier to use buildozer with WSL, as documented in the code, it works around a windows filesystem limitation that prevents updates of android tools/platorm-tools.

Needs more testing, but i could build with it after a clean/disclean, without manual workarounds.

### DIFF
--- a/buildozer/targets/android.py
+++ b/buildozer/targets/android.py
@@ -11,6 +11,9 @@ import sys
 if sys.platform == 'win32':
     raise NotImplementedError('Windows platform not yet working for Android')
 
+from platform import uname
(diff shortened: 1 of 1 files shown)


buildozer/targets/android.py
```diff
@@ -22,6 +25,7 @@
 import io
 import re
 import ast
+import sh
```
AndreMiras: then we may want to add the `sh` dependency to the setup.py requirement

tshirtman: indeed!

PR #741, opened 2018-11-09
Explicitly run `pip3` when using Python 3

dbrnz:
This resolved issues when both Python 2 and Python 3 were installed.

### DIFF
--- a/buildozer/targets/android.py
+++ b/buildozer/targets/android.py
@@ -11,6 +11,8 @@ import sys
 if sys.platform == 'win32':
     raise NotImplementedError('Windows platform not yet working for Android')
 
+PIP_COMMAND = 'pip' if (sys.version_info[0] < 3) else 'pip3'
+
 ANDROID_API = '19'
 ANDROID_MINAPI = '9'
 ANDROID_SDK_VERSION = '20'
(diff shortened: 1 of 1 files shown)


buildozer/targets/android.py
```diff
@@ -535,7 +537,7 @@ def _install_p4a(self):
         options = "--user"
         if "VIRTUAL_ENV" in os.environ or "CONDA_PREFIX" in os.environ:
             options = ""
-        cmd('pip install -q {} {}'.format(options, " ".join(pip_deps)))
+        cmd('{} install -q {} {}'.format(PIP_COMMAND, options, " ".join(pip_deps)))
```
AndreMiras: Or would it make sense to `python -m pip install -q ...` instead? I think that would pick the correct pip module based on the Python version running.
That way we could skip that `PIP_COMMAND = 'pip if...` thing above. What do you think?

AndreMiras:
Great, I actually also had the issue recently, but from what I recall I tried hardcoding `pip3` to give it a try, but it didn't seem enough somehow.
I need to give it another try before merging.
What about `python -m pip install...` instead of setting `pip3` explicitly?

dbrnz:
I did wonder about `python -m pip` and now think this would be the best way to go, so long as the `python` used is the one running `buildozer` (e.g. one may want `python3 -m pip`). I'm also thinking of the case where say there is both `python35` and `python36` on the PATH and buildozer has been installed into say python35. Mind you anybody doing this should be using a virtual environment...

AndreMiras:
So you just gave it a try with `python -m pip` and it behaves as expected @dbrnz ?

dbrnz:
To be honest, I'm in the process of doing so at the moment but won't really be able to put much into this until later today (NZ time).

dbrnz:
All good now and tested under Ubuntu 18:
```
Cloning into 'python-for-android-new-toolchain'...
# Run "/usr/bin/python3 -m pip install -q --user 'appdirs' 'colorama>=0.3.3' 'jinja2' 'six'"
```

PR #743, opened 2018-11-10
Correctly write out a sub-process's `stdout` when using Python 3.

dbrnz:
`stderr` was handled correctly but not `stdout`...

### DIFF
--- a/buildozer/__init__.py
+++ b/buildozer/__init__.py
@@ -311,7 +311,7 @@ class Buildozer(object):
                     ret_stdout.append(chunk)
                 if show_output:
                     if IS_PY3:
-                        stdout.write(str(chunk))
+                        stderr.write(chunk.decode('utf-8'))
                     else:
                         stdout.write(chunk)
             if fd_stderr in readx:


buildozer/__init__.py
```diff
@@ -311,7 +311,7 @@ def cmd(self, command, **kwargs):
                     ret_stdout.append(chunk)
                 if show_output:
                     if IS_PY3:
-                        stdout.write(str(chunk))
+                        stderr.write(chunk.decode('utf-8'))
```
AndreMiras: well I only see this now as I'm working for https://github.com/kivy/buildozer/issues/857
But I think we should not be writing to `stderr` here right? Feels like a failed copy paste

dbrnz: Indeed...

AndreMiras: Thanks for confirming I was still doubting. I'm working on a unit test now for #857 it's almost done so I'll address that as well.

AndreMiras: Addressed in https://github.com/kivy/buildozer/pull/905

PR #747, opened 2018-11-17
fix ndk URLs

zworkb:
buildozer failed with downloading recent versions of android NDK.
This diff reflects the changes in the android NDK download urls.
I tested it with the recent NDKs (16b,17c).


tshirtman:
doh, missed that earlier when i did #754, sorry, seems to be a better fix, we should probably rebase/merge it anyway.

PR #757, opened 2018-11-30
Update cython version from the docs

AndreMiras:
`cython==0.28.6` seems to be doing just fine on p4a until now.
https://github.com/kivy/python-for-android/blob/df60061/Dockerfile#L120


inclement:
Good catch, thanks.

PR #759, opened 2018-12-01
allow specifying of Android NDK API. Resolves #758

### LINKED ISSUE #758: Need to be able to specify the NDK API target
With Google Play target API level changing, it's impossible to have buildozer create an android apk that allows minimum api of 19. 

Need the ability to specify the android NDK api and pass that to p4a, so this type of buildozer configuration can be supported:

<pre>
# (int) Android API to use
android.api = 27

# (int) Minimum API required
android.minapi = 19

# (int) Android SDK version to use
android.sdk = 23

# (str) Android NDK version to use
android.ndk = 17c

# (int) Android NDK API to use (optional)
android.ndk.api = 19

</pre>

brentpicasso:
This will let us specify the NDK API, which will allow us to create an .apk that has a min API of 19 (or lower?) while still allowing the targeting of higher API levels, like 27. 

Adds a new entry to the buildozer spec as well.  Resolves #758

### DIFF
--- a/buildozer/default.spec
+++ b/buildozer/default.spec
@@ -99,7 +99,7 @@ fullscreen = 0
 # (str) Android NDK version to use
 #android.ndk = 9c
 
-# (int) Android NDK API to use (optional). This is the minimum API your API will support. 
+# (int) Android NDK API to use (optional). This is the minimum API your app will support. 
 #android.ndk.api = 19
 
 # (bool) Use --private data storage (True) or --dir public storage (False)
--- a/buildozer/default.spec
+++ b/buildozer/default.spec
@@ -90,7 +90,7 @@ fullscreen = 0
 # (int) Android API to use
 #android.api = 19
 
-# (int) Minimum API required
+# (int) Minimum API required. You will need to set the android.ndk.api to be as low as this value.
 #android.minapi = 9
 
 # (int) Android SDK version to use
@@ -99,7 +99,7 @@ fullscreen = 0
 # (str) Android NDK version to use
 #android.ndk = 9c
 
-# (int) Android NDK API to use (optional)
+# (int) Android NDK API to use (optional). This is the minimum API your API will support. 
 #android.ndk.api = 19
 
 # (bool) Use --private data storage (True) or --dir public storage (False)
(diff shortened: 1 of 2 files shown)


brentpicasso:
@inclement  This is the issue and PR we discussed in Discord chat. Thanks for the help earlier!

buildozer/default.spec
```diff
@@ -99,6 +99,9 @@ fullscreen = 0
 # (str) Android NDK version to use
 #android.ndk = 9c
 
+# (int) Android NDK API to use (optional)
```
inclement: Could you add something like "This is the minimum API that your APK will support.", I think that will be helpful to people.

Could also add that note with the minsdk line, i.e. that you must set ndk_api to also be as low as the minsdk.

brentpicasso: Ok, made the change. please check if I got what you meant. Thanks!

buildozer/default.spec
```diff
@@ -99,6 +99,9 @@ fullscreen = 0
 # (str) Android NDK version to use
 #android.ndk = 9c
 
+# (int) Android NDK API to use (optional). This is the minimum API your API will support.
```
inclement: API is repeated (second one should be app), but other than that great! :)

brentpicasso: Whoops!  Too much multi-tasking. Ok, fixed.

buildozer/default.spec
```diff
@@ -99,6 +99,9 @@ fullscreen = 0
 # (str) Android NDK version to use
 #android.ndk = 9c
 
+# (int) Android NDK API to use (optional). This is the minimum API your app will support. 
+#android.ndk.api = 19
```
inclement: Just one more minor thing, could you make it `android.ndk_api` rather than `android.ndk.api`? I think the former is more consistent with the other spec lines. Sorry about being picky, but the spec api becomes rather set in stone once release because people start depending on it, so it's worth making sure to get it right.

brentpicasso: Done!

inclement:
Thanks again!

buildozer/default.spec
```diff
@@ -90,7 +90,7 @@ fullscreen = 0
 # (int) Android API to use
 #android.api = 19
 
-# (int) Minimum API required
+# (int) Minimum API required. You will need to set the android.ndk.api to be as low as this value.
```
AndreMiras: `s/android.ndk.api/android.ndk_api/

Issue #769, opened 2018-12-19
'# Android NDK is missing, downloading' - Android NDK is installed


drjoms:
### Versions

* Python:
python --version
Python 3.6.5
eselect  python list
Available Python interpreters, in order of preference:
  [1]   python3.6
  [2]   python2.7
  [3]   python3.4
  [4]   python3.5 (fallback)

* OS:
uname -a
Linux (none)dimko's Desktop 4.17.0-gentoo #6 SMP PREEMPT Sat Dec 8 22:29:46 GMT 2018 x86_64 AMD Ryzen 5 1600 Six-Core Processor AuthenticAMD GNU/Linux

* Kivy:
dev-python/Kivy
      Latest version available: 1.10.1
      Latest version installed: 1.10.1

* Cython:
  dev-python/cython
      Latest version available: 0.25.2
      Latest version installed: 0.25.2

### Description
 /home/dimko/.local/bin/buildozer android debug deploy run
# Check configuration tokens
# Ensure build layout
# Check configuration tokens
# Preparing build
# Check requirements for android
# Install platform
# Apache ANT found at /home/dimko/.buildozer/android/platform/apache-ant-1.9.4
# Android SDK found at /home/dimko/.buildozer/android/platform/android-sdk-20
# Android NDK is missing, downloading
Traceback (most recent call last):
  File "/home/dimko/.local/bin/buildozer", line 11, in <module>
    sys.exit(main())
  File "/home/dimko/.local/lib64/python3.6/site-packages/buildozer/scripts/client.py", line 13, in main
    Buildozer().run_command(sys.argv[1:])
  File "/home/dimko/.local/lib64/python3.6/site-packages/buildozer/__init__.py", line 1059, in run_command
    self.target.run_commands(args)
  File "/home/dimko/.local/lib64/python3.6/site-packages/buildozer/target.py", line 92, in run_commands
    func(args)
  File "/home/dimko/.local/lib64/python3.6/site-packages/buildozer/target.py", line 102, in cmd_debug
    self.buildozer.prepare_for_build()
  File "/home/dimko/.local/lib64/python3.6/site-packages/buildozer/__init__.py", line 176, in prepare_for_build
    self.target.install_platform()
  File "/home/dimko/.local/lib64/python3.6/site-packages/buildozer/targets/android.py", line 507, in install_platform
    self._install_android_ndk()
[...]

PR #773, opened 2018-12-25
- #769 - fixed ndk download a bit, refactored dl fun

mahomahomaho:
ndk download dir for ndk9 is /android/ndk not /android/repository/


guysoft:
I think this can be closed because #747  was merged

AndreMiras:
Indeed and thanks @mahomahomaho for the PR attempt

PR #775, opened 2018-12-27
Docker from current git

mahomahomaho:
it's nice to build docker with current, bleeding edge or even uncommited version of buildozer, instead of using only stable, pypi pushed buildozer version. 

In such case, --build-arg git=true should be used.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -44,10 +44,10 @@ setup(
     license='MIT',
     packages=[
         'buildozer', 'buildozer.targets', 'buildozer.libs', 'buildozer.scripts'
-    ],
+        ],
     package_data={'buildozer': ['default.spec']},
     include_package_data=True,
-    install_requires=['pexpect', 'virtualenv'],
+    install_requires=['pexpect', 'virtualenv', 'sh'],
     classifiers=[
         'Development Status :: 4 - Beta', 'Intended Audience :: Developers',
         'Topic :: Software Development :: Build Tools',
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,18 +1,28 @@
 # Dockerfile for providing buildozer
 # Build with:
 # docker build --tag=buildozer .
+# 
+# if you want to use bleeding edge/current version from git, instead from pypi, then use
+# docker build --tag=buildozer --build-arg git=true .
+#
 # In order to give the container access to your current working directory
 # it must be mounted using the --volume option.
 # Run with (e.g. `buildozer --version`):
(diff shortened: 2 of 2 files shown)


mahomahomaho:
Yes, I'm Lukasz Mach and I don't know why it took my Name/Surname instead of nick. Probably infected with my other settings. 

Re commit - how can I remove it? This commit is based on #773, I believe that once #773 is merged to master, then this commit will disappear from list of commits and from diff.

KeyWeeUsr:
Ok, just wanted the confirmation. About that commit, you can do

    git add .
    git stash
    git reset HEAD~2
    git add *Dockerfile* *setup.py*
    git commit -m "<some msg>"
    git push -f
    git stash pop

to save all your unsaved changes to stash, remove the two commits (but leave the changes available), add only the Docker related changes, create a clean commit, force push to this branch (you have a different branch for PR #773) and retrieve back the saved changes from stash.

Also, don't add unnecessary spaces on this line https://github.com/kivy/buildozer/pull/775/commits/7158ceb3d36631e1ac7055257d5c3205743a9263#diff-2eeaed663bd0d25b7e608891384b7298R47.

Otherwise the (docker related) changes are fine and I'd merge them. I'm not sure about NDK changes, that's why I want to remove the previous commit :)

mahomahomaho:
ok, so I did like you said. It's not exactly clear what do those commands do, but apparently they work.

setup.py
```diff
@@ -44,10 +44,10 @@ def find_version(*file_paths):
     license='MIT',
     packages=[
         'buildozer', 'buildozer.targets', 'buildozer.libs', 'buildozer.scripts'
-    ],
+        ],
```
KeyWeeUsr: Could you please remove this empty change?

AndreMiras:
> I wonder if the code that uses sh was only recently included or whether it just silently failed because p4a included that dependency too.

Yes @KeyWeeUsr the `sh` "regression" was introduced in https://github.com/kivy/buildozer/pull/740

> ok, so I did like you said. It's not exactly clear what do those commands do, but apparently they work.

@mahomahomaho yes git squashing and rebasing looks awkward at first, but it's actually a pretty awesome thing to know and use

setup.py
AndreMiras: Yes I also got bugged by it, but I tried to not say anything because I'm always the one nit picking :stuck_out_tongue: 
So sorry if I saw your comment before I would not have merged :confused:

Issue #778, opened 2018-12-30
Outdated Android NDK download URL in Kivy Virtual Machine


stroobandt:
Using the latest Version 2.0, released the 13 May 2017, of the Kivy virtual Machine
and after having updated `buildozer` to version `0.37` as required by the instructions,
the `buildozer` build process fails when trying to unzip `android-ndk-r9c-linux-x86_64.tar.bz2`

This happens because on old URL to the Android NDK is being referenced, namely:
<http://dl.google.com/android/repository/android-ndk-r9c-linux-x86_64.tar.bz2>

The issue can be circumvented by manually downloading the Android NDK using `wget` to the `~/.buildozer/android/platform/` directory but this time from the correct URL, namely:
<http://dl.google.com/android/ndk/android-ndk-r9c-linux-x86_64.tar.bz2>

This issue was [previously reported by other people on StackExchange](https://stackoverflow.com/a/42191418/2192488).

brady-ds:
Based on [caver456's comment](https://github.com/kivy/buildozer/issues/760#issuecomment-445544664), would this be the same as #760?

guysoft:
This also happends in master, Yes it looks like #760, same error here

guysoft:
Fixed in master, this issue can be closed

inclement:
Thanks @guysoft, we also now default to a much newer NDK.

PR #767, opened 2018-12-12
Fixed download url for ndk versions < 11

jtoledo1974:
The URL for ndk versions < 11 doesn't work anymore.
I've found a complete listing for direct links and verified that they work here http://pnsurez.blogspot.com/2015/07/download-android-ndk-tools.html
It was troublesome because the default 9c ndk wasn't downloading.

Issue #760, opened 2018-12-01
Buildozer Error `ValueError: read of closed file`


TheDeaX-Lab:
### Versions

* Python: 3.7.1
* OS: ArchLinux 4.19.4-arch1-1-ARCH
* Kivy: 1.10.1
* Cython: 0.29.1

### Description

I compile my project "Hello world" and catch the error on the program Buildozer

### buildozer.spec
[buildozer.txt](https://github.com/kivy/buildozer/files/2635809/buildozer.txt)

### Logs
```
# Check configuration tokens
# Ensure build layout
# Check configuration tokens
# Preparing build
# Check requirements for android
# Install platform
# Apache ANT found at /home/deax/.buildozer/android/platform/apache-ant-1.9.4
# Android SDK found at /home/deax/.buildozer/android/platform/android-sdk-20
# Android NDK is missing, downloading
Traceback (most recent call last):
  File "/usr/bin/buildozer", line 11, in <module>
    sys.exit(main())
  File "/usr/lib/python3.7/site-packages/buildozer/scripts/client.py", line 13, in main
    Buildozer().run_command(sys.argv[1:])
  File "/usr/lib/python3.7/site-packages/buildozer/__init__.py", line 1059, in run_command
    self.target.run_commands(args)
  File "/usr/lib/python3.7/site-packages/buildozer/target.py", line 92, in run_commands
    func(args)
  File "/usr/lib/python3.7/site-packages/buildozer/target.py", line 102, in cmd_debug
    self.buildozer.prepare_for_build()
  File "/usr/lib/python3.7/site-packages/buildozer/__init__.py", line 176, in prepare_for_build
    self.target.install_platform()
  File "/usr/lib/python3.7/site-packages/buildozer/targets/android.py", line 507, in install_platform
    self._install_android_ndk()
  File "/usr/lib/python3.7/site-packages/buildozer/targets/android.py", line 319, in _install_android_ndk
    cwd=self.buildozer.global_platform_dir)
  File "/usr/lib/python3.7/site-packages/buildozer/__init__.py", line 687, in download
    urlretrieve(url, filename, report_hook)
  File "/usr/lib/python3.7/urllib/request.py", line 1817, in retrieve
    block = fp.read(bs)
  File "/usr/lib/python3.7/tempfile.py", line 481, in func_wrapper
    return func(*args, **kwargs)
ValueError: read of closed file
```

codingsett:
[...]

PR #780, opened 2019-01-07
Fixes ndk url in current buildozer #778 #760  #767

guysoft:
Currently kivy will not build any modern app, unless the url is fixed.

This has been stuck for a while. And reported many times. See #778 #760  #767
Also see: https://stackoverflow.com/questions/42097586/buildozer-android-ndk-not-downloading-ubuntu

There also seem to be pull requests about this that have been ignored, so I suggest if this pull request is not merged, please direct me how to make it so it would get merged and people people following kivy tutorials don't reach dead ends.

### DIFF
--- a/buildozer/targets/android.py
+++ b/buildozer/targets/android.py
@@ -313,7 +313,7 @@ class TargetAndroid(Target):
         unpacked = 'android-ndk-r{0}'
         archive = archive.format(self.android_ndk_version, architecture)
         unpacked = unpacked.format(self.android_ndk_version)
-        url = 'http://dl.google.com/android/repository/'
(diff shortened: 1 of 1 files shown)


AndreMiras:
So you're suggesting to revert part of https://github.com/kivy/buildozer/pull/754 ?
Could you please align with @tshirtman who introduced related changes recently

tito:
The changes would broke recent ndk, and works only with old one. Actually, the PR #747 was ok, i just fixed it and it should be ok now.

guysoft:
Yes, #747 is better, thanks for merging it!

PR #787, opened 2019-01-12
Specify GitHub branches in buildozer.spec

hackalog:
Currently (for an iOS build) buildozer downloads 2 platform build dependencies from github repos:
* kivy-ios
* ios-deploy

This change allows the repo/branch of those dependencies to be specified in `buildozer.spec`, either as local (custom) checkouts, or by specifying a git URL and branch. In fact, it allows any git-based dependency to be handled in the same manner, either by specifying a local checkout directory to use, or by specifying a URL/branch pair.

This has been essential for me in order to lock the versions of upstream dependencies during periods of high code churn (or to refer my build to forked, bug-fixed repos of these upstream dependencies)

`buildozer.spec` can now be used to specify the location of these repos as follows:

```
# Uncomment the following line to use the local check-out to install `kivy-ios`
# If specified, this overrides any choice of url/branch below
# ios.kivy_ios_dir = ../kivy-ios

# uncomment the following lines to specify a custom (forked) kivy-ios repo/branch
# specifying no branch will default to using master
ios.kivy_ios_url = https://github.com/learnleapfly/kivy-ios
ios.kivy_ios_branch = fix_py2
```

This mechanism is platform agnostic, so I have moved the helper functions to be methods of `Target` rather than `TargetIos`.  I'm not sure if p4a handles platform dependencies by checking out git repos at present.

### DIFF
--- a/buildozer/default.spec
+++ b/buildozer/default.spec
@@ -220,6 +220,16 @@ android.arch = armeabi-v7a
 
 # (str) Path to a custom kivy-ios folder
 #ios.kivy_ios_dir = ../kivy-ios
+# Alternately, specify the URL and branch of a git checkout:
+ios.kivy_ios_url = https://github.com/kivy/kivy-ios
+ios.kivy_ios_branch = master
+
+# Another platform dependency: ios-deploy
+# Uncomment to use a custom checkout
+#ios.ios_deploy_dir = ../ios_deploy
+# Or specify URL and branch
+ios.ios_deploy_url = https://github.com/phonegap/ios-deploy
+ios.ios_deploy_branch = 1.7.0
 
 # (str) Name of the certificate to use for signing the debug version
 # Get a list of available identities: buildozer ios list_identities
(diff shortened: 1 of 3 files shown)


AndreMiras:
Well done for that PR :muscle:  I like the idea of being able to choose the repo url.
However I'm wondering are we not overengineering things with configuring the url_format and all of that. Can't it simply be git repo url and branch, that's all we would need isn't it?

hackalog:
Owner is maybe superfluous. I could take that out. 

Repo is used both to construct the buildozer key and in the URL, and branch is used both in the URL and in the git clone command, so I'd want to keep those

hackalog:
(the url alone isn't sufficient, as you'd have to specify the buildozer key in addition, so I don't know that you gain anything that way)

AndreMiras:
Awesome thanks! It looks good to me. It would be great if @tito could take a look since he really knows about ios :pray:

tito:
Very nice PR, well done!

Issue #782, opened 2019-01-07
Buildozer 0.37 UnicodeDecodeError: 'ascii' codec can't decode byte 0xc2 


homdx:
I use this env in docker provided by @JonasT
https://github.com/JonasT/p4a-build-spaces

./p4aspaces shell p4a-py3-api28ndk21

#buildozer init
```
Traceback (most recent call last):
  File "/usr/local/bin/buildozer", line 11, in <module>
    sys.exit(main())
  File "/usr/local/lib/python3.6/dist-packages/buildozer/scripts/client.py", line 13, in main
    Buildozer().run_command(sys.argv[1:])
  File "/usr/local/lib/python3.6/dist-packages/buildozer/__init__.py", line 122, in __init__
    self.config.read(filename)
  File "/usr/lib/python3.6/configparser.py", line 697, in read
    self._read(fp, filename)
  File "/usr/lib/python3.6/configparser.py", line 1015, in _read
    for lineno, line in enumerate(fp, start=1):
  File "/usr/lib/python3.6/encodings/ascii.py", line 26, in decode
    return codecs.ascii_decode(input, self.errors)[0]
UnicodeDecodeError: 'ascii' codec can't decode byte 0xc2 in position 1665: ordinal not in range(128)
```
Buildozer 0.37

AndreMiras:
Looks like https://github.com/kivy/buildozer/pull/789 can you confirm @homdx ?

homdx:
Great!!! It's are fixed
Thanks to @guysoft

```
./p4aspaces shell p4a-py3-api28ndk21
Sending build context to Docker daemon  7.168kB
Step 1/31 : FROM ubuntu
 ---> 1d9c17228a9e
Step 2/31 : ENV SDK_TOOLS="sdk-tools-linux-4333796.zip"
 ---> Using cache
 ---> 304f2f432153
Step 3/31 : ENV NDK_DL="https://dl.google.com/android/repository/android-ndk-r17c-linux-x86_64.zip"
 ---> Using cache
 ---> addbf4f66f6c
Step 4/31 : ENV NDKVER=r17c
 ---> Using cache
 ---> 8e9d71d8b0fa
Step 5/31 : ENV NDKDIR=/ndk/
 ---> Using cache
 ---> 4892c51ff6e3
Step 6/31 : ENV NDKAPI=21
 ---> Using cache
 ---> 6c5180f8ab9f
Step 7/31 : ENV ANDROIDAPI=28
[...]

PR #789, opened 2019-01-13
Handle buildozer.spec with unicode chars

guysoft:
Other wise you get a UnicodeEncodeError in python3.

### DIFF
--- a/buildozer/__init__.py
+++ b/buildozer/__init__.py
@@ -119,7 +119,7 @@ class Buildozer(object):
         self.config.getrawdefault = self._get_config_raw_default
 
         if exists(filename):
-            self.config.read(filename)
+            self.config.read(filename, "utf-8")
             self.check_configuration_tokens()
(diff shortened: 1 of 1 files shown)


guysoft:
The reason this is needed, because of the line:
https://github.com/kivy/buildozer/blob/master/buildozer/default.spec#L65
```
author = © Copyright Info
```

AndreMiras:
Thanks for the fix, that may fix https://github.com/kivy/buildozer/issues/782
Funny to see it pops only now. Also I'm not sure why I can't reproduce it myself :thinking:

homdx:
@guysoft , thank you!

Log:
https://github.com/kivy/buildozer/issues/782#issuecomment-453915678

hackalog:
Note for posterity. This caused a python2 regression. Fixed in https://github.com/kivy/buildozer/pull/799

PR #791, opened 2019-01-16
Fix python 2 crash when reading config file

JonasT:
Fix python 2 crash when reading config file, also see https://github.com/kivy/buildozer/commit/c66af5dc0bf7420df6d71c26bee55f085b80c93b#commitcomment-31965387

### DIFF
--- a/buildozer/__init__.py
+++ b/buildozer/__init__.py
@@ -119,7 +119,10 @@ class Buildozer(object):
         self.config.getrawdefault = self._get_config_raw_default
 
         if exists(filename):
-            self.config.read(filename, "utf-8")
+            try:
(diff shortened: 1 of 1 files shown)


AndreMiras:
Nice, that was fast. And I'm not using python2. Even though it would be easy to try in Python2 I think this bug is perfect to push continuous integration testing, see https://github.com/kivy/buildozer/issues/679

hackalog:
Works for me. Please pull this, as buildozer is completely broken on python2 otherwise

AndreMiras:
ahah I was on this actually tito, I wanted to finish with the regression test. Will do in 5 minutes

Issue #790, opened 2019-01-13
Add support for --frameworks on iOS


tito:
See https://github.com/kivy/kivy-ios/pull/339

hackalog:
I've got a WIP for buildozer: https://github.com/kivy/buildozer/pull/795 that would implement this.

The WIP relies on a slight improvement to the framework support in kivy-ios, namely https://github.com/kivy/kivy-ios/pull/342

PR #795, opened 2019-01-17
Support the --add-frameworks flag in kivy-ios

hackalog:
This makes use of the newly added `--add-frameworks` flag in kivy-ios to add frameworks named in
`ios.frameworks` in `buildozer.spec`

Note, this is a WIP until https://github.com/kivy/kivy-ios/pull/342 is accepted (which this change relies on).

This would fix https://github.com/kivy/buildozer/issues/790


hackalog:
I've dropped the WIP, but this is still blocked on https://github.com/kivy/kivy-ios/pull/342 getting pulled

PR #802, opened 2019-01-26
Show envs when run2

mahomahomaho:
When some command fails, it's good to know what envs was set, to be able to reproduce error. 

Also few formatting changes forced by flake8

### DIFF
--- a/buildozer/__init__.py
+++ b/buildozer/__init__.py
@@ -11,7 +11,6 @@ __version__ = '0.38.dev0'
 import os
 import re
 import sys
-import zipfile
 import select
 import codecs
 import textwrap
@@ -24,6 +23,9 @@ from os import environ, unlink, walk, sep, listdir, makedirs
 from copy import copy
 from shutil import copyfile, rmtree, copytree, move
 from fnmatch import fnmatch
+
+from pprint import pformat
+
 try:
     from urllib.request import FancyURLopener
     from configparser import SafeConfigParser
@@ -41,7 +43,7 @@ try:
     colorama.init()
 
     RESET_SEQ = colorama.Fore.RESET + colorama.Style.RESET_ALL
-    COLOR_SEQ = lambda x: x
+    COLOR_SEQ = lambda x: x  # noqa: E731
     BOLD_SEQ = ''
     if sys.platform == 'win32':
         BLACK = colorama.Fore.BLACK + colorama.Style.DIM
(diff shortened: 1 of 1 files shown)


buildozer/__init__.py
```diff
@@ -335,6 +346,7 @@ def cmd(self, command, **kwargs):
         process.communicate()
         if process.returncode != 0 and break_on_error:
             self.error('Command failed: {0}'.format(command))
+            self.log_env(1, kwargs['env'])
```
AndreMiras: Thanks @mahomahomaho I think that PR makes sense. I have a couple of minor questions however.
Why is the log level hard-coded to two different values, once with `2` and then here with `1`?
Also why aren't we consistent in the way to access `env` key in kwargs? Once with `kwargs.get("env")` and then with `kwargs['env']`?

AndreMiras:
I think that makes sense, thank you @mahomahomaho 
Just added a couple of comments, could you review?

buildozer/__init__.py
mahomahomaho: yes, you're right. It should be kwargs['env'] in both cases, because we're sure that kwargs contains 'env' key.

mahomahomaho: regarding two different levels - It's because in this place, you will get info in the same level as msg aabove (error). 

I was thinking about various approachs, and I considered to define log_debug_env and log_info_env ... but I decided to not make too many methods. But indeed - 1 and 2 could be somehow ENUMed.

AndreMiras:
Thanks for taking another look. I just realised that yes that hardcoded number things was already in our code base :sweat_smile: I'll unit test some of it, refactor and probably merge your PR after further review

AndreMiras:
I've created https://github.com/kivy/buildozer/pull/831 to help with testing the logger before we can merge yours

buildozer/__init__.py
```diff
@@ -282,6 +296,7 @@ def cmd(self, command, **kwargs):
             else:
                 self.debug('Run {0!r} ...'.format(command.split()[0]))
         self.debug('Cwd {}'.format(kwargs.get('cwd')))
+        self.log_env(self.DEBUG, kwargs["env"])
```
AndreMiras: I think that like should be removed.
I'm seeing it only now after using buildozer master for a while.
For instance during the auto accept license process, this `log_env` seems to be printing every time. We want it only after crashes, not for every command call

AndreMiras: I've created a PR to remove that line https://github.com/kivy/buildozer/pull/843

PR #803, opened 2019-01-30
Update for p4a master

inclement:
With these changes, buildozer defaults to using python-for-android's master branch, and with the next defaults will compile successfully.


inclement:
Tested with python2 and python3, merging.

PR #806, opened 2019-02-01
Removed android_old and updated SDK management to work with current SDK

inclement:
Seems to work but could do with some testing in different situations.


buildozer/targets/android.py
```diff
@@ -248,30 +306,30 @@ def _install_android_sdk(self):
...
-            archive = 'android-sdk_r{0}-windows.zip'
+            archive = 'sdk-tools-windows-4333796.zip'
             unpacked = 'android-sdk-windows'
         elif platform in ('darwin', ):
-            archive = 'android-sdk_r{0}-macosx.zip'
+            archive = 'sdk-tools-darwin-4333796.zip'
             unpacked = 'android-sdk-macosx'
         elif platform.startswith('linux'):
-            archive = 'android-sdk_r{0}-linux.tgz'
+            archive = 'sdk-tools-linux-4333796.zip'
```
AndreMiras: Minor: this `4333796` part should probably be a const somewhere

buildozer/targets/android.py
```diff
@@ -420,68 +520,54 @@ def _install_android_packages(self):
...
-        if not os.access(self.android_cmd, os.X_OK):
-            self.buildozer.cmd('chmod ug+x {}'.format(self.android_cmd))
-
         # 1. update the tool and platform-tools if needed
-        packages = self._android_list_sdk()
-        skip_upd = self.buildozer.config.getdefault('app',
-                                                    'android.skip_update', False)
-        if 'tools' in packages or 'platform-tools' in packages:
-            if not skip_upd:
-                if WSL:
```
AndreMiras: breaking change, but I don't really mind about that one :trollface:

buildozer/targets/android.py
```diff
@@ -586,81 +678,190 @@ def _install_p4a(self):
         options = "--user"
         if "VIRTUAL_ENV" in os.environ or "CONDA_PREFIX" in os.environ:
             options = ""
+        print('pa dir', self.pa_dir)
```
tito: debugging?

inclement:
Merged!

For general information, this *will* cause buildozer to download a new SDK for every single user unless they already manually specify the SDK directory, as the old sdk version parameter is now ignored. It would be possible to work around this and use the old directory where availabel, but instead I just removed the parameter entirely to minimise legacy code. This has the big advantage that it will ensure everyone is using newer build tools with all the android APIs available, which would otherwise be an issue.

inclement:
@gottadiveintopython Thanks for pointing it out, should be fixed by https://github.com/kivy/buildozer/pull/807

You will need to update to a newer SDK though, android-sdk-20 is too old and now not supported due to lacking sdkmanager.

gottadiveintopython:
no problem :-) thanks for the fix

well I use python3. According to [this](https://github.com/kivy/kivy/wiki/Android-SDK-NDK-Information#working-sdkndk-kgs---known-good-set), I thought SDK 20 is the only choice for python3.

inclement:
@gottadiveintopython That page is out of date, I will update it. Any recent SDK is fine, if you let buildozer download the SDK it will pull something appropriate.

gottadiveintopython:
oh ... didnt know that. maybe thats why I always struggle with building apk.
Thanks!

buildozer/targets/android.py
```diff
@@ -341,43 +406,85 @@ def _install_android_ndk(self):
...
-        child = self.buildozer.cmd_expect(
-            '{} update sdk -u -a -t {}'.format(
-                self.android_cmd,
-                packages,
-                cwd=self.buildozer.global_platform_dir),
-            timeout=None,
-            env={
+
+        auto_accept_license = self.buildozer.config.getdefault(
+            'app', 'android.accept_sdk_license', False)
```
AndreMiras: there's a bug documented in https://github.com/kivy/buildozer/issues/816

PR #808, opened 2019-02-03
Update README.md to add opencollective

tito:


AndreMiras:
It's not clear to me why the build is still failing. Can you take a look @tito ?

Issue #816, opened 2019-02-05
android.accept_sdk_license may misbehave


AndreMiras:
### Versions

* Python:
* OS:
* Buildozer: master 205f1f6

### Description

I suspect `android.accept_sdk_license` misbehave such that if it's defined to `False` in the spec file like `android.accept_sdk_license = False`. It will be evaluated as `True` still.
This is because the `auto_accept_license = self.buildozer.config.getdefault('app', 'android.accept_sdk_license', False)` expression retrieve the string and would return `"False"` which is evaluated as `True` in bool, see: https://github.com/kivy/buildozer/blob/a780cc4/buildozer/targets/android.py#L463

Demo:
`buildozer.spec` must contain `android.accept_sdk_license = False`
Then from a shell:
```python
>>> from buildozer import Buildozer
>>> buildozer = Buildozer()
# Check configuration tokens
>>> buildozer.config.get('app', 'android.accept_sdk_license')
'False'
>>> bool('False')
True
```

AndreMiras:
PR https://github.com/kivy/buildozer/pull/817

AndreMiras:
Fixed via https://github.com/kivy/buildozer/pull/817

PR #817, opened 2019-02-05
Use getbooldefault() for boolean, fixes #806

AndreMiras:


inclement:
Thanks, I thought about this flaw but thought getdefault handled it :o

AndreMiras:
The commit messages refs #806 when it should have been #816 :grimacing:

Issue #813, opened 2019-02-04
Running buildozer help crashes


AndreMiras:
### Versions

* Python: python3
* OS: Ubuntu 18.04
* Kivy: N/A
* Cython: N/A
* buildozer master e0d29f9

### Description

Running `buildozer help` command crashes.
I suspect it's a bug introduced in our recent refactoring https://github.com/kivy/buildozer/pull/806
Buildozer was installed with:
```
pip3 install --upgrade https://github.com/kivy/buildozer/archive/e0d29f9.zip
```
The error is:
```
TypeError: join() argument must be str or bytes, not 'NoneType'
```
See detailed log and trace below.

### buildozer.spec
Command:
```sh
buildozer help
```

Spec file:
N/A

### Logs
```
$ buildozer help
# Check configuration tokens
Usage:
    buildozer [--profile <name>] [--verbose] [target] <command>...
    buildozer --version

Available targets:
  android            Android target, based on python-for-android project (old toolchain)

Global commands (without target):
  distclean          Clean the whole Buildozer environment.
  help               Show the Buildozer help.
  init               Create a initial buildozer.spec in the current directory
  serve              Serve the bin directory via SimpleHTTPServer
  setdefault         Set the default command to run when no arguments are given
  version            Show the Buildozer version

Target commands:
  clean      Clean the target environment
  update     Update the target dependencies
  debug      Build the application in debug mode
  release    Build the application in release mode
  deploy     Deploy the application on the device
  run        Run the application on the device
  serve      Serve the bin directory via SimpleHTTPServer
Traceback (most recent call last):
  File "/home/andre/.local/bin/buildozer", line 11, in <module>
    load_entry_point('buildozer==0.38.dev0', 'console_scripts', 'buildozer')()
  File "/home/andre/.local/lib/python3.6/site-packages/buildozer/scripts/client.py", line 13, in main
    Buildozer().run_command(sys.argv[1:])
  File "/home/andre/.local/lib/python3.6/site-packages/buildozer/__init__.py", line 1034, in run_command
    getattr(self, cmd)(*args)
  File "/home/andre/.local/lib/python3.6/site-packages/buildozer/__init__.py", line 1095, in cmd_help
    self.usage()
  File "/home/andre/.local/lib/python3.6/site-packages/buildozer/__init__.py", line 976, in usage
    mt = m.get_target(self)
  File "/home/andre/.local/lib/python3.6/site-packages/buildozer/targets/android.py", line 1231, in get_target
[...]

PR #821, opened 2019-02-08
Unit tests buildozer --help command, refs #813

AndreMiras:
This test should fail due to #813. Next commit will fix it.

### DIFF
--- a/buildozer/targets/android.py
+++ b/buildozer/targets/android.py
@@ -1225,4 +1225,5 @@ class TargetAndroid(Target):
 
 
 def get_target(buildozer):
+    buildozer.targetname = "android"
     return TargetAndroid(buildozer)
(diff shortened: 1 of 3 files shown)


AndreMiras:
CI build of the tests without fix is failing as expected, see https://travis-ci.com/kivy/buildozer/builds/100271525
Subsequent commit should fix it

AndreMiras:
Second commit fixes the build as expected https://travis-ci.com/kivy/buildozer/builds/100272034

buildozer/targets/android.py
```diff
@@ -1225,4 +1225,5 @@ def cmd_logcat(self, *args):
 
 
 def get_target(buildozer):
+    buildozer.targetname = "android"
```
AndreMiras: I would say it's pretty bad practice and misleading mutating a get parameter, but this is how it was handled in the past

PR #831, opened 2019-02-17
Unit test logger

AndreMiras:
- adds test utils methods
- checks default log level
- sets log level via spec file
- checks messages printing depending on level

### DIFF
--- a/tests/test_buildozer.py
+++ b/tests/test_buildozer.py
@@ -1,12 +1,58 @@
+import re
 import os
+import codecs
 import mock
 import unittest
 import buildozer as buildozer_module
 from buildozer import Buildozer
 from six import StringIO
+import tempfile
 
 
 class TestBuildozer(unittest.TestCase):
+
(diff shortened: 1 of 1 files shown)


tests/test_buildozer.py
```diff
@@ -1,13 +1,59 @@
...
 from buildozer import Buildozer
 from six import StringIO
+import tempfile
 
 
 class TestBuildozer(unittest.TestCase):
 
+    def setUp(self):
+        """
+        Creates a temporary spec file containing the content of the default.spec.
```
AndreMiras: note: I know this is not the best approach performance wise since it creates a different temp file for each tests. But well it automates per-test spec file isolation and I don't think it impacts unit tests performances for now anyway

AndreMiras:
This PR request is one more step toward better code coverage and it will help in merging https://github.com/kivy/buildozer/pull/802 (after rebasing)

Issue #823, opened 2019-02-10
download() is not CI log friendly


AndreMiras:
The way we report download progress in [`buildozer/__init__.py` download()](https://github.com/kivy/buildozer/blob/91ba12d/buildozer/__init__.py#L672) is absolutely helpful in the terminal, but really messes up with Travis logger.

When downloading things as big as the Android NDK via buildozer, it produces a very large amount of progress lines, like in the extract below:
```
[0m[1;34m# Android NDK is missing, downloading[0m
- Download 0.00%
- Download 0.00%
- Download 0.00%
- Download 0.00%
- Download 0.00%
- Download 0.01%
- Download 0.01%
- Download 0.01%
- Download 0.01%
- Download 0.01%
- Download 0.01%
- Download 0.01%
- Download 0.01%
- Download 0.02%
- Download 0.02%
- Download 0.02%
```
In total this log file has over 100K lines of `- Download `! See attached screenshot and complete log for reference: https://api.travis-ci.org/v3/job/491113892/log.txt
![image](https://user-images.githubusercontent.com/24973/52528534-506b8c00-2ce1-11e9-97e1-fc5026a8354b.png)

To fix the issue, maybe we should make it possible to reduce the verbosity of this `download()` method. Other suggestions are very welcome.
Side note: same issue exists in p4a via `download_file()`, but the most annoying currently is `buildozer` as it generates lager downloads so far.

mkg20001:
My fork, which includes [this commit](https://github.com/mkg20001/buildozer/commit/c4ae19f4b8bcd24db1dcd053bb175c0bdaa06d60) has an environment flag called `CI_MODE` which can be set to 1 to disable progress-printing

AndreMiras:
Very nice, yes that's the idea :+1:

mkg20001:
Created PRs #833 and kivy/python-for-android#1727 that implement this. Hope this gets merged soon, then I can finally drop my forks of this project

PR #833, opened 2019-02-24
Add ci_mode to toggle download progress

mkg20001:
Ref #823

### DIFF
--- a/buildozer/__init__.py
+++ b/buildozer/__init__.py
@@ -692,8 +692,9 @@ class Buildozer(object):
             else:
                 progression = '{0:.2f}%'.format(
                         index * blksize * 100. / float(size))
-            stdout.write('- Download {}\r'.format(progression))
-            stdout.flush()
+            if "CI" not in environ:
+                stdout.write('- Download {}\r'.format(progression))
+                stdout.flush()
 
         url = url + filename
         if cwd:


AndreMiras:
Great work! Same comment as https://github.com/kivy/python-for-android/pull/1727#issuecomment-466819110 could you review/address? :pray:

mkg20001:
Noticed that the unzip command still spams

screenshot

Would anybody mind looking into that?

AndreMiras:
Merged thanks!
Regarding the extraction command I think it comes from here https://github.com/kivy/python-for-android/blob/6660b9e/pythonforandroid/recipe.py#L405 but I haven't verified yet.
Basically `tar -tf` the `-t` flag is the verbose part, see man page:
> List the contents of an archive.  Arguments are optional.  When given, they specify the names of the members to list.

But let's keep it for another story so I'm merging your PR which already is an improvement. Thanks for making a commit amend/squash by the way, I also like clean commit log :smile:

mkg20001:
Could you merge https://github.com/kivy/python-for-android/pull/1727 as well, please?

AndreMiras:
Yes but for https://github.com/kivy/python-for-android/pull/1727 to make sense we need the .travis.yml update as well, do you mind making the change to the PR? :pray:

Issue #837, opened 2019-02-26
`sdkmanager` updates spamming logs in continuous integration


mkg20001:
### Versions

* Python: Python 2.7.15rc1
* OS: Ubuntu 18.04 amd64
* Buildozer: Buildozer 0.40.dev0 (modified fork)

### Description

Currently during CI runs `sdkmanager` install/update commands create HUGE amounts of log output https://gitlab.com/mkg20001-gh/ZeroNet-kivy/-/jobs/167339825/raw

### buildozer.spec
Command:
```sh
buildozer android release
```

Spec file:
```
[app]
warn_on_root = 0
buildozer.warn_on_root = 0
title = ZeroNet
package.name = zeronet
package.domain = net.mkg20001
source.dir = src
source.include_exts = py,png,jpg,kv,atlas
version.regex = self\.version = ['"](.*)['"]
version.filename = %(source.dir)s/zero/src/Config.py
requirements = sqlite3,typing,openssl,m2crypto,gevent,msgpack-python,pil,hostpython2,git+https://github.com/kivy/plyer,android,kivy
p4a.branch = master
presplash.filename = %(source.dir)s/splash.png
icon.filename = %(source.dir)s/icon.png
orientation = all
services = zn:service.py
osx.python_version = 3
osx.kivy_version = 1.9.1
fullscreen = 0
android.permissions = INTERNET
android.api = 28
android.minapi = 21
android.ndk_api = 21
android.private_storage = True
android.whitelist = sqlite3
android.accept_sdk_license = True
p4a.branch = master
android.arch = armeabi-v7a
ios.kivy_ios_url = https://github.com/kivy/kivy-ios
ios.kivy_ios_branch = master
ios.ios_deploy_url = https://github.com/phonegap/ios-deploy
ios.ios_deploy_branch = 1.7.0
[buildozer]
log_level = 2
warn_on_root = 1
```

### Logs

https://gitlab.com/mkg20001-gh/ZeroNet-kivy/-/jobs/167339825/raw

More than 40MB

mkg20001:
Found a fix, will PR soon

mkg20001:
PR #838

Julian-O:
This PR appears to have stalled. 

@mkg20001: Are you still active? 

Does someone else want to have a go?

PR #838, opened 2019-02-26
feat: disable showing sdk manager progress

mkg20001:
It has no quiet mode and causes between 10 and 100 MEGABYTES of log 
spam. I mean c'mon.

So let's just discard all of that using a simple grep vanish

### DIFF
--- a/buildozer/targets/android.py
+++ b/buildozer/targets/android.py
@@ -99,6 +99,10 @@ class TargetAndroid(Target):
         """Call the sdkmanager in our Android SDK with the given arguments."""
         # Use the android-sdk dir as cwd by default
         kwargs['cwd'] = kwargs.get('cwd', self.android_sdk_dir)
+        if "CI" in environ and ">/dev/null" not in args:
+            args = list(args)
+            args.insert(len(args), ' | grep -v "^\\["')
+            args = tuple(args)
         command = self.sdkmanager_path + ' ' + ' '.join(args)
         return_child = kwargs.pop('return_child', False)
         if return_child:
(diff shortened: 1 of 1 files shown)


mkg20001:
Fixes #837

buildozer/targets/android.py
```diff
@@ -467,6 +471,10 @@ def _android_update_sdk(self, *sdkmanager_commands):
             command = '{} | {} --licenses'.format(
                 yes_command, self.sdkmanager_path)
             self.buildozer.cmd(command, cwd=self.android_sdk_dir)
+        if "CI" in environ:
+            sdkmanager_commands = list(sdkmanager_commands)
+            sdkmanager_commands.insert(len(sdkmanager_commands), '>/dev/null')  # mute stdout, important stuff goes to stderr anyways
+            sdkmanager_commands = tuple(sdkmanager_commands)
```
AndreMiras: why not skipping this whole part since you already introduced the `grep -v "^\\["` fix in the `_sdkmanager()` method above?

buildozer/targets/android.py
```diff
@@ -99,6 +99,10 @@ def _sdkmanager(self, *args, **kwargs):
         """Call the sdkmanager in our Android SDK with the given arguments."""
         # Use the android-sdk dir as cwd by default
         kwargs['cwd'] = kwargs.get('cwd', self.android_sdk_dir)
+        if "CI" in environ and ">/dev/null" not in args:
+            args = list(args)
+            args.insert(len(args), ' | grep -v "^\\["')
+            args = tuple(args)
         command = self.sdkmanager_path + ' ' + ' '.join(args)
```
AndreMiras: Instead if this 4 lines of code "adding" the `grep` o the args. I think it would be more simple to go after the `command = self.sdkmanager_path + ' ' + ' '.join(args)` line and to do something like:
```
if 'CI' in environ:
    # hide download lines which look like "[=          ] 10% Downloading platform-tools"
    command += r' | grep -v "^\["'
```
Also note that I've used a raw-string to make the grep expression more readable (by skipping the double escaping of the `\`).
Last I've added a little code comment because it's not super obvious what we are doing

PR #848, opened 2019-03-08
Various Dockerfile improvements

AndreMiras:
- installs cutting edge version from current source
- removes outdated java version fixes
- migrates to host Python3
- verifies the image builds and run via Travis

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,9 +2,6 @@
 # Build with:
 # docker build --tag=buildozer .
 # 
-# if you want to use bleeding edge/current version from git, instead from pypi, then use
-# docker build --tag=buildozer --build-arg git=true .
-#
 # In order to give the container access to your current working directory
 # it must be mounted using the --volume option.
 # Run with (e.g. `buildozer --version`):
@@ -13,16 +10,12 @@
 # docker run --volume "$(pwd)":/home/user/hostcwd --entrypoint /bin/bash -it --rm buildozer
 
 FROM ubuntu:18.04
-ARG git
 
-ENV BDOZER_REQ=${git:+"/src"}
-ENV BDOZER_REQ=${BDOZER_REQ:-"buildozer"}
 ENV USER="user"
 ENV HOME_DIR="/home/${USER}"
 ENV WORK_DIR="${HOME_DIR}/hostcwd" \
+    SRC_DIR="${HOME_DIR}/src" \
     PATH="${HOME_DIR}/.local/bin:${PATH}"
-
-COPY . /src
 
 # configures locale
 RUN apt update -qq > /dev/null && \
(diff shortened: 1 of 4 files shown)


buildozer/__init__.py
```diff
@@ -511,9 +511,9 @@ def check_application_requirements(self):
 
         # did we already installed the libs ?
         if exists(self.applibs_dir) and \
-           self.state.get('cache.applibs', '') == requirements:
-                self.debug('Application requirements already installed, pass')
-                return
+               self.state.get('cache.applibs', '') == requirements:
+            self.debug('Application requirements already installed, pass')
+            return
```
AndreMiras: I wish I didn't have to touch that, but for some reason (flake8 update?) the linter started complaining with:
```
buildozer/__init__.py:515:17: E117 over-indented
```
Also the diff is a bit messed up, but basically I'm simply changing from:
```python
        # did we already installed the libs ?
        if exists(self.applibs_dir) and \
           self.state.get('cache.applibs', '') == requirements:
                self.debug('Application requirements already installed, pass')
                return
```
to:
```python
        # did we already installed the libs ?
        if exists(self.applibs_dir) and \
               self.state.get('cache.applibs', '') == requirements:
            self.debug('Application requirements already installed, pass')
            return
```
So block stay the same, just fixing the over/under indentation

AndreMiras:
The long term idea is not only to use Docker for testing purpose, but eventually to completely replace the VMs. For now I want to keep the feature as undocumented as I'm still unsure if that Docker image will serve both purposes or if we would prefer to have a dedicated `buildozer-docker` repository for it.
Basically if we really want to make the buildozer container useful for people to use, we need to ship it with a lot more system dependencies so it can compile most of the recipes. e.g. `cmake autoconf libltdl-dev gettext autogen libtool pkg-config libffi-dev zip automake`
Still I think this pull request is a nice iteration, because not only it fixes the currently broken Docker image, it also makes sure it won't ever be broken by making it part of the continuous integration.

buildozer/__init__.py
tshirtman: i tend to prefer this form
```python
        if (
            exists(self.applibs_dir)
            and self.state.get('cache.applibs', '') == requirements
        ):
            self.debug('Application requirements already installed, pass')
            return
```
which i think is easier to read, but not a big deal.

Dockerfile
```diff
@@ -35,14 +28,14 @@ ENV LANG="en_US.UTF-8" \
...
+    sudo python3-pip python3-setuptools file
 
 # https://buildozer.readthedocs.io/en/latest/installation.html#android-on-ubuntu-18-04-64bit
 RUN dpkg --add-architecture i386 && apt update -qq > /dev/null && \
 	apt install -qq --yes --no-install-recommends \
 	build-essential ccache git libncurses5:i386 libstdc++6:i386 libgtk2.0-0:i386 \
-	libpangox-1.0-0:i386 libpangoxft-1.0-0:i386 libidn11:i386 python2.7 \
-	python2.7-dev openjdk-8-jdk unzip zlib1g-dev zlib1g:i386
+	libpangox-1.0-0:i386 libpangoxft-1.0-0:i386 libidn11:i386 python2.7 python3 \
+	python3-dev openjdk-8-jdk unzip zlib1g-dev zlib1g:i386
```
tshirtman: i didn't even install the i386 architecture in any of my recent build, i think this is completly irrelevant now, but i might be missing something, anyway, that can probably be part of another change.

AndreMiras: Good point, it may be worth to look at in subsequent pull requests.
I actually want to discuss some approaches for Docker

buildozer/__init__.py
AndreMiras: Thanks addressed!
Just the `and` part had to stay at the end of the line or `flake8` would complain

tshirtman: ah, i guess either the version is not up to date with spec, or we have a config that goes against the current recommendations, but we can live with that :).

AndreMiras: I remember that rule changed recently indeed.
Once I'm doing providing a fully functional replacement for our VMs, I'll probably resume my work on fixing most of the linting

PR #856, opened 2019-03-13
Fix for bug when logcat value which cannot be decode to utf-8

Draqun:
Hi.

I spend 2 days to detect problem with debugging program on my Xiaomi device. So I propose this simple patch which allow replace character which cannot be decode to UTF-8.

Best regards.
Draqun


AndreMiras:
Thanks for the PR :pray: !
It would have been perfect with a unit test :sweat_smile:

Draqun:
Hi.
Now I don't have time to do it, but in free time in next month I'll try do it.
Best regards.
Draqun

PR #860, opened 2019-03-16
Download Apache ANT at the same path as the buildozer.spec android.ant_path option

robertpro:
ANT installation path was hardcoded so if you changed the installation path, it will download every time you run `buildozer android XXX`
```
# Android ANT is missing, downloading
# Downloading http://archive.apache.org/dist/ant/binaries/apache-ant-1.9.4-bin.tar.gz
# Run 'tar xzf apache-ant-1.9.4-bin.tar.gz'
# Cwd /home/roberto/.buildozer/android/platform
# Apache ANT installation done.
```

I am sorry, I did not found any easy way to test this change since it will download ANT which we don't want that on testing

### DIFF
--- a/buildozer/targets/android.py
+++ b/buildozer/targets/android.py
@@ -290,14 +290,17 @@ class TargetAndroid(Target):
             self.buildozer.info('Apache ANT found at {0}'.format(ant_dir))
             return ant_dir
 
+        if not os.path.exists(ant_dir):
+            os.makedirs(ant_dir)
+
         self.buildozer.info('Android ANT is missing, downloading')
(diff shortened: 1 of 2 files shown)


AndreMiras:
Thanks that sounds relevant.
Thanks for mentioning tests, by the way, yes there's probably a way to test it with `unittest.mock` if you `mock.patch` both `buildozer.download()` and `buildozer.file_extract()` and verify it was called with the expected params.

Anyway the PR looks fine, but I still want to give it a try before merging. In the meantime if you want to give unit tests a try, feel free :smile:

robertpro:
Thanks @AndreMiras, I learned something new :) please review!

inclement:
Thanks @robertpro, sorry for the delay merging this.

PR #932, opened 2019-07-03
Doc/installation updates

tshirtman:
instructions for android were quite outdated, added a troubleshooting section and also instructions for kivy-ios, following a documentation from cruor99

### DIFF
--- a/docs/source/installation.rst
+++ b/docs/source/installation.rst
@@ -46,3 +46,23 @@ Buildozer didn't install a necessary package
     ~/.buildozer/android/platform/android-sdk/tools/bin/sdkmanager --install "build-tools;29.0.0"
 
 Then press "y" then enter to accept the license.
+
+
+Targeting IOS
+~~~~~~~~~~~~~
+
+Install XCode and command line tools (through the AppStore)
+
+
+Install homebrew (https://brew.sh)
+
+::
+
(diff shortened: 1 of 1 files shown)


docs/source/installation.rst
```diff
@@ -13,26 +13,56 @@ First, install the buildozer project with::
...
 
-If you target Android, you must install at least Cython, few build libs, and a
-Java SDK. Some binaries of the Android SDK are still in 32 bits, so you need
-few 32bits libraries available:
+Android on Ubuntu (64bit)
+~~~~~~~~~~~~~~~~~~~~~~~~~
 
-Android on Ubuntu 18.04 (64bit)
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+(tested on 18.04, 18.10, 19.04)
```
AndreMiras: I would rather just keep the LTS title `18.04`, because it's the one that will truly be under CI not the other.
I mean maybe you or somebody tested 18.10, 19.04 but not on regular basis like a CI would do.
So the idea is not that we don't support other version, but more to say that we're confident it works for a given one

tshirtman: right, it's true that it's not going to be tested continuously, so it might break in the future., i'll edit.

PR #940, opened 2019-07-24
New feature: allow to use a p4a fork

opacam:
This PR adds the ability to use any p4a fork. So combined with buildozer.spec file key `p4a.branch` it will allows us to test any pull requests of `kivy/python-for-android` using buildozer. Also, an user could use his own fork/branch of p4a editing two keys from his buildozer.spec file: `p4a.fork` and `p4a.branch` to make use of his own fork/branch of p4a.

So...in this PR we do:
  - add a new key to buildozer.spec file: `p4a.fork`
  - check on each build that fork/branch has not been changed, and if so, remove the old p4a installation (which will trigger a new p4a's installation)
  - move key `p4a.branch` to proper section (python-for-android)...of course this is not mandatory, it's only to be neatest and to make easier to find the key
  - fix a pep8 error (two blank lines before class)

**Note:** The p4a fork it's supposed to have the same name than the the repository from kivy's organization (python-for-android), if that is not the case, don't expect this to work!!!

### DIFF
--- a/buildozer/default.spec
+++ b/buildozer/default.spec
@@ -158,9 +158,6 @@ fullscreen = 0
 # (list) Java classes to add as activities to the manifest.
 #android.add_activites = com.example.ExampleActivity
 
-# (str) python-for-android branch to use, defaults to master
-#p4a.branch = master
-
 # (str) OUYA Console category. Should be one of GAME or APP
 # If you leave this blank, OUYA support will not be enabled
 #android.ouya.category = GAME
@@ -207,6 +204,9 @@ android.arch = armeabi-v7a
 #
 # Python for android (p4a) specific
 #
+
+# (str) python-for-android branch to use, defaults to master
+#p4a.branch = master
 
 # (str) python-for-android git clone directory (if empty, it will be automatically cloned from github)
 #p4a.source_dir =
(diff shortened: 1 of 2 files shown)


buildozer/default.spec
```diff
@@ -208,6 +205,12 @@ android.arch = armeabi-v7a
 # Python for android (p4a) specific
 #
 
+# (str) python-for-android fork to use, defaults to upstream (kivy)
+#p4a.fork = kivy
```
AndreMiras: Minor remark: so for now, we're assuming the fork is on github only right?

AndreMiras:
Awesome thanks! I'll give this a try later this week hopefully

buildozer/default.spec
opacam: Yes, and also we assume that the name of the fork has not changed :wink:

AndreMiras: That's more than enough for a first iteration and that probably covers more than 80% of the cases

AndreMiras:
Oh man, I'm just using it now for testing https://github.com/kivy/python-for-android/pull/1883
This is such an improvement to me. It will also really ease people hooking to their forks until PRs get merged upstream.
Well done :clap:

PR #947, opened 2019-08-01
Add ability to get p4a's recommended android's NDK version

opacam:
So this way, if an user has not specified the android's NDK version via buildozer.spec's file, We will try to get the recommended one by p4a, established in `pythonforandroid.recommendations.py` file, so depending on the version of p4a used, we can use an android NDK or another.

This will allow us to solve the problem where p4a needs to move to a newer android NDK in develop branch, but also we need to preserve the old one for p4a's master branch (like the `ndk-r19-migration` thing that we are dealing this days in p4a, kivy/python-for-android#1722).

**Note:** in case that an old version of p4a is used, previous to 2019, where the p4a's recommendations file still not existed, We will fallback to buildozer's hardcoded NDK version

### DIFF
--- a/buildozer/targets/android.py
+++ b/buildozer/targets/android.py
@@ -47,6 +47,12 @@ DEPRECATED_TOKENS = (('app', 'android.sdk'), )
 # does.
 DEFAULT_SDK_TAG = '4333796'
 
+MSG_P4A_RECOMMENDED_NDK_ERROR = (
+    "WARNING: Unable to find recommended android's NDK for current "
+    "installation of p4...so returning the recommended buildozer's one, which "
+    "is android's NDK r{android_ndk}".format(android_ndk=ANDROID_NDK_VERSION)
+)
+
 
 class TargetAndroid(Target):
     targetname = 'android'
@@ -54,6 +60,7 @@ class TargetAndroid(Target):
     p4a_fork = 'kivy'
     p4a_branch = 'master'
     p4a_apk_cmd = "apk --debug --bootstrap="
+    p4a_best_ndk_version = None
     extra_p4a_args = ''
 
     def __init__(self, *args, **kwargs):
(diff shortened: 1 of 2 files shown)


AndreMiras:
I haven't deeply checked yet, but that sounds like a very cool idea :clap: 
Could you also add the unit tests for this new feature?

opacam:
Oh yes, I will try to do something to cover this :wink:

opacam:
Ok, added unittests,
this should cover almost all the lines I added :wink:

AndreMiras:
Awesome, thanks for that! I'll try reviewing later. I still have a lot to catch up with on other PR, but I have to say PR with tests usually get higher prio for me :smile: So you may have yours reviewed earlier than expected

buildozer/targets/android.py
```diff
@@ -97,6 +104,47 @@ def _p4a(self, cmd, **kwargs):
...
+        if not hasattr(self, "pa_dir"):
+            pa_dir = join(self.buildozer.platform_dir, self.p4a_directory)
+        else:
+            pa_dir = self.pa_dir
+
+        # check p4a's recommendation file, and in case that exists find the
+        # recommended android's NDK version, otherwise return buildozer's one
+        ndk_version = ANDROID_NDK_VERSION
+        rec_file = join(pa_dir, "pythonforandroid", "recommendations.py")
+        if (not os.path.exists(pa_dir)) or (not os.path.isfile(rec_file)):
```
AndreMiras: Minor: why not only checking for `os.path.isfile(rec_file)`?

opacam: Thanks @AndreMiras :smile:

opacam:
thanks @AndreMiras :smile:

since it seems that @inclement has no opposition on this one (kivy/python-for-android#1962)

I merge this :+1:

PR #950, opened 2019-08-07
added cmake to requirements

mcroni:


AndreMiras:
Thanks for the PR.
I'm not to sure what to think about this. Because that actually isn't a buildozer dependency, but a p4a recipe dependency

PR #967, opened 2019-09-08
Rename final apk with arch in the name

tito:
If i do 2 build with the same buildozer spec, but 2 profiles with different arch (or 2 buildozer spec), i don't want the final apk to be overwritten. So now the final apk name contains the arch:

For example:
- `connectage-0.4.0.23-arm64-v8a-release.apk`
- `connectage-0.4.0.23-armeabi-v7a-release.apk`

### DIFF
--- a/buildozer/targets/android.py
+++ b/buildozer/targets/android.py
@@ -1127,8 +1127,10 @@ class TargetAndroid(Target):
             apk = u'{packagename}-{mode}.apk'.format(
                 packagename=packagename, mode=mode)
             apk_dir = join(dist_dir, "build", "outputs", "apk", mode_sign)
-            apk_dest = u'{packagename}-{version}-{mode}.apk'.format(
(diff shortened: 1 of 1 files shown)


buildozer/targets/android.py
```diff
@@ -1127,8 +1127,10 @@ def build_package(self):
             apk = u'{packagename}-{mode}.apk'.format(
                 packagename=packagename, mode=mode)
             apk_dir = join(dist_dir, "build", "outputs", "apk", mode_sign)
-            apk_dest = u'{packagename}-{version}-{mode}.apk'.format(
-                packagename=packagename, mode=mode, version=version)
+
+            apk_dest = u'{packagename}-{version}-{arch}-{mode}.apk'.format(
```
AndreMiras: We could move it outside the if/else and have that part made only once since we're doing exactly the same inside and outside the if/else

PR #969, opened 2019-09-14
Add libs only for current arch

misl6:
When multiple `android.add_libs_*` are present, buildozer is trying (and failing) to copy libs also for non-current arch.

This PR goal is to leave the `buildozer.spec` file unchanged when targeting multiple archs except for the `android.arch` option.


AndreMiras:
Thank you for the PR! I've never played with that feature, but I think that PR makes sense.
I still need to give it a quick try. In the meantime, if you come up with a way to unit test it, that would for sure help in the future.

PR #970, opened 2019-09-15
fix: show output of sdk update if auto_accept_license is false

### LINKED ISSUE #916: Accept license terms prompt is not visible
buildozer version: 0.39
python : 3.7

When log level is low or `--verbose` is false the license accept prompt is not visible in the shell, and it seems that the process is stuck. The last line is `# Installing/updating SDK platform tools if necessary`

When --verbose is added, the long license text becomes visible, but there is still no prompt in the end, asking for user input - you just have to guess that you have to type 'y' and hit enter.

related: #66 and #870

tshirtman:
untested, but should fix #916

### DIFF
--- a/buildozer/targets/android.py
+++ b/buildozer/targets/android.py
@@ -515,6 +515,7 @@ class TargetAndroid(Target):
         auto_accept_license = self.buildozer.config.getbooldefault(
             'app', 'android.accept_sdk_license', False)
 
+        kwargs = {}
         if auto_accept_license:
             # `SIGPIPE` is not being reported somehow, but `EPIPE` is.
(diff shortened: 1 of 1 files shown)


AndreMiras:
Good one, thanks.
I'm willing to test it before merge, but I can't find a way to revert my setup to status prior accepting license. And deleting the entire `~/.buildozer` directory feels a bit overkill :stuck_out_tongue:

tshirtman:
I was actually testing it just now, and it seem to work :) i got the license displayed and i could confirm it manually.

tshirtman:
ok, it does work :)

tshirtman:
yay \o/

PR #983, opened 2019-10-05
Fixes test_p4a_recommended_android_ndk_found() mocking

AndreMiras:
Mocking `for line in open()` should be done via `StringIO`.
Also uses `pytest` in `tox.ini` so assert errors are more verbose.

### DIFF
--- a/tests/test_buildozer.py
+++ b/tests/test_buildozer.py
@@ -7,7 +7,10 @@ from buildozer import Buildozer, IS_PY3
 from six import StringIO
 import tempfile
 
-import mock
+try:
+    from unittest import mock  # Python 3
+except ImportError:
+    import mock  # Python 2
 
 from buildozer.targets.android import (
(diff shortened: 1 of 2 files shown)


tests/test_buildozer.py
```diff
@@ -7,7 +7,10 @@
 from six import StringIO
 import tempfile
 
-import mock
+try:
+    from unittest import mock  # Python 3
+except ImportError:
+    import mock  # Python 2
```
AndreMiras: I know you guys like it this way, so I reintroduced it :smile:

tox.ini
```diff
@@ -2,9 +2,10 @@
 envlist = pep8,py27,py3
 
 [testenv]
-deps = mock
-commands =
-    python -m unittest discover --top-level-directory=. --start-directory=tests/
+deps =
+    mock
+    pytest
+commands = pytest tests/
```
AndreMiras: Basically `pytest` makes it possible to see exact result of failing comparison. That way we don't need to debug print around in Travis :wink:

Issue #1052, opened 2020-04-08
ImportError: cannot import name 'main'


brvier:
Build failed while trying to install python dependancies :

```
[INFO]:    *** PYTHON PACKAGE / PROJECT INSTALL STAGE ***
[INFO]:    The requirements (certifi, gpxpy, oscpy, plyer) don't have recipes, attempting to install them with pip
[INFO]:    If this fails, it may mean that the module has compiled components and needs a recipe.
[INFO]:    -> directory context /home/user/hostcwd/.buildozer/android/platform/build-arm64-v8a/build
stty: 'standard input': Inappropriate ioctl for device
[INFO]:    -> running virtualenv --python=python3 venv
Exception in thread background thread for pid 38291:                                               
Traceback (most recent call last):
  File "/usr/lib/python3.6/threading.py", line 916, in _bootstrap_inner
    self.run()
  File "/usr/lib/python3.6/threading.py", line 864, in run
    self._target(*self._args, **self._kwargs)
  File "/home/user/.local/lib/python3.6/site-packages/sh.py", line 1540, in wrap
    fn(*args, **kwargs)
  File "/home/user/.local/lib/python3.6/site-packages/sh.py", line 2459, in background_thread
    handle_exit_code(exit_code)
  File "/home/user/.local/lib/python3.6/site-packages/sh.py", line 2157, in fn
    return self.command.handle_command_exit_code(exit_code)
  File "/home/user/.local/lib/python3.6/site-packages/sh.py", line 815, in handle_command_exit_code
    raise exc
sh.ErrorReturnCode_1: 

  RAN: /home/user/.local/bin/virtualenv --python=python3 venv

  STDOUT:
Traceback (most recent call last):
  File "/home/user/.local/bin/virtualenv", line 7, in <module>
    from virtualenv import main
ImportError: cannot import name 'main'


  STDERR:

Traceback (most recent call last):
  File "/usr/lib/python3.6/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "/usr/lib/python3.6/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/home/user/hostcwd/.buildozer/android/platform/python-for-android/pythonforandroid/toolchain.py", line 1212, in <module>
    main()
  File "/home/user/hostcwd/.buildozer/android/platform/python-for-android/pythonforandroid/entrypoints.py", line 18, in main
    ToolchainCL()
[...]

PR #1053, opened 2020-04-08
[Python3] #1052 ImportError: cannot import name 'main'

Sirfanas:
Original Issue is: [ImportError: cannot import name 'main' #1052](https://github.com/kivy/buildozer/issues/1052)

This is the fix I show on [this post](https://github.com/kivy/buildozer/issues/1052#issuecomment-611006248)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -75,6 +75,8 @@ WORKDIR ${WORK_DIR}
 COPY --chown=user:user . ${SRC_DIR}
 
 # installs buildozer and dependencies
-RUN pip3 install --user Cython==0.28.6 ${SRC_DIR}
+RUN pip3 install --user wheel Cython==0.28.6 ${SRC_DIR}
+
+RUN python3 -m pip install -U pip virtualenv
 
 ENTRYPOINT ["buildozer"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -77,6 +77,7 @@ COPY --chown=user:user . ${SRC_DIR}
 # installs buildozer and dependencies
 RUN pip3 install --user wheel Cython==0.28.6 ${SRC_DIR}
 
-RUN python3 -m pip install -U pip virtualenv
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -75,6 +75,8 @@ WORKDIR ${WORK_DIR}
 COPY --chown=user:user . ${SRC_DIR}
 
 # installs buildozer and dependencies
-RUN pip3 install --user Cython==0.28.6 ${SRC_DIR}
+RUN pip3 install --user wheel Cython==0.28.6 ${SRC_DIR}
+
+RUN python3 -m pip install -U pip virtualenv
```
AndreMiras, line 80: Thanks for the PR, and why the inconsistency one time with `pip3 install` and the other time `python3 -m pip install`?

Sirfanas: Just a reflex I have to use `python3 -m pip`
I change the line into 'pip3'

brvier:
I confirm also the workarround.

Dockerfile
```diff
@@ -75,6 +75,9 @@ WORKDIR ${WORK_DIR}
 COPY --chown=user:user . ${SRC_DIR}
 
 # installs buildozer and dependencies
-RUN pip3 install --user Cython==0.28.6 ${SRC_DIR}
+RUN pip3 install --user wheel Cython==0.28.6 ${SRC_DIR}
+
+# Upgrade virtualenv and pip
+RUN pip3 install -U pip virtualenv
```
AndreMiras, line 81: Thanks for the update. How about in-lining the entire thing?
```suggestion
RUN pip3 install --user --upgrade Cython==0.28.6 wheel pip virtualenv ${SRC_DIR}
```

Sirfanas: Yeah look really better, I just build the image and compile my app, it's ok for me !

PR #1059, opened 2020-04-10
Setup coverage testing, closes #1058

AndreMiras:
Also enables report to coveralls.io via GitHub Action.
Note the `COVERALLS_REPO_TOKEN` was setup via:
https://github.com/kivy/buildozer/settings/secrets
And is accessible for repos admin via:
https://coveralls.io/github/kivy/buildozer

PR #1060, opened 2020-04-10
Setup coverage testing, closes #1058

### LINKED ISSUE #1058: Setup coverage testing
We want to have an idea of the project test coverage. Increasing the coverage should help avoiding trivial issues like https://github.com/kivy/buildozer/pull/1045
Ideally the coverage test would be hooked with an online service like https://coveralls.io/

AndreMiras:
Also enables report to coveralls.io via GitHub Action.
Note the `COVERALLS_REPO_TOKEN` was setup via:
https://github.com/kivy/buildozer/settings/secrets
And is accessible for repos admin via:
https://coveralls.io/github/kivy/buildozer

This is a follow up for https://github.com/kivy/buildozer/pull/1059 branching from upstream to have access to the secrets

### DIFF
--- a/.github/workflows/test_python.yml
+++ b/.github/workflows/test_python.yml
@@ -25,7 +25,7 @@ jobs:
 
     - name: Requirements
       run: |
-        pip install --timeout=120 -U setuptools tox>=2.0
+        pip install -U coveralls setuptools tox>=2.0
         tox
     - name: Build
       run: |
@@ -33,3 +33,7 @@ jobs:
     - name: Test
       run: |
         docker run buildozer --version
+    - name: coveralls
+      run: coveralls
+      env:
+        COVERALLS_REPO_TOKEN: ${{ secrets.COVERALLS_REPO_TOKEN }}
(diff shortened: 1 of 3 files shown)


AndreMiras:
Build is actually green but github as some leftover from the previous PR I think.
It's not ideal as it doesn't upload coveralls for PRs but only upon merge. But we have at least the coverage result from tox for PRs.
image

AndreMiras:
Thanks for the check. Yes I'd say let's see, if it always break I would try to fix it

AndreMiras:
Weird I realise I don't see one change I did on this PR :thinking: 
I've probably messed up somewhere, basically the change was to run the coverage only on push to master, that way the `secrets.COVERALLS_REPO_TOKEN` is only pulled when allowed.
Without it pull requests would probably fail. I'll see soon as I'm about to pull request now. If it fails then I'll bring the fix

AndreMiras:
Just confirmed and fixed via https://github.com/kivy/buildozer/pull/1062

PR #1068, opened 2020-04-13
Dedicated Docker build/run job

AndreMiras:
Creates a dedicated job for Docker testing so it can run in parallel with the other tests.
Prior this change the CI was running in 3 minutes, it now runs in 2.


AndreMiras:
Yes I was wondering the same, thanks for sharing, I think that would make sense too.
I'm not against a PR, I'm merging this one since it's already one step toward more isolation.
Thanks for the review

PR #1069, opened 2020-04-13
Unit tests TargetAndroid.build_package()

AndreMiras:
Increases android.py coverage from 24% to 36% and total coverage from 25% to 29%.
Note that this tests would have caught the glob issue reported in #1044.

### DIFF
--- a/tests/targets/test_android.py
+++ b/tests/targets/test_android.py
@@ -45,6 +45,12 @@ def patch_buildozer_checkbin():
     return patch_buildozer("checkbin")
 
 
+def patch_target_android(method):
+    return mock.patch(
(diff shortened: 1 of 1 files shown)


tests/targets/test_android.py
```diff
@@ -41,10 +41,20 @@ def patch_os_isfile():
     return mock.patch("os.path.isfile")
 
 
+def patch_os_listdir():
+    return mock.patch("os.listdir")
+
+
```
opacam, line 47: Nice, 4.9% more of coverage :smile:

**Note:** I think that we never use this mock, right?...are you planing to use it later?

AndreMiras: Yes good catch thanks. I actually started using it and decided no to for now.
I've removed it now

PR #1078, opened 2020-04-24
Also tests against macOS platform

AndreMiras:
Increases coverage from 28% to 30%.

### DIFF
--- a/tests/targets/test_android.py
+++ b/tests/targets/test_android.py
@@ -51,6 +51,10 @@ def patch_target_android(method):
     )
 
 
+def patch_platform(platform):
+    return mock.patch("buildozer.targets.android.platform", platform)
+
+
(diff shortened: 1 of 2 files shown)


tests/targets/test_android.py
```diff
@@ -146,7 +153,8 @@ def test_check_configuration_tokens(self):
             self.target_android.check_configuration_tokens()
         assert m_check_configuration_tokens.call_args_list == [mock.call([])]
 
-    def test_install_android_sdk(self):
+    @pytest.mark.parametrize("platform", ["linux", "darwin",])
```
opacam, line 156: **nitpick:** In this case, I think that I would remove the trailing comma

AndreMiras: Addressed

PR #1083, opened 2020-05-02
Integration testing

AndreMiras:
Checks:
- buildozer installs
- buildozer --help
- buildozer init

Currently `buildozer android debug` is not ran as it's not yet fully reliable.

### DIFF
--- a/.github/workflows/test_python.yml
+++ b/.github/workflows/test_python.yml
@@ -17,10 +17,10 @@ jobs:
 
     runs-on: ${{ matrix.os }}
     steps:
-    - uses: actions/checkout@master
+    - uses: actions/checkout@v2
 
     - name: Setup python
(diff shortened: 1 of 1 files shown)


.github/workflows/test_python.yml
```diff
@@ -38,8 +38,23 @@ jobs:
...
     - name: Docker build
       run: docker build --tag=kivy/buildozer .
     - name: Docker run
       run: docker run kivy/buildozer --version
+
+  Integration:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Setup python
+      uses: actions/setup-python@v1
```
opacam, line 51: **nitpick:** we could use `actions/setup-python@v2`

AndreMiras: Addressed :+1:

PR #1094, opened 2020-05-08
Drops Python 2 support

AndreMiras:
Makes sure the user is warned at install time if using Python 2.


AndreMiras:
Thanks @opacam yes I've increased the margin to 1% (p4a is 2%), let's see if we can hold it :)

PR #1098, opened 2020-05-11
#1097 Add toml, pep517 and colorama to fix "android clean" command

Sirfanas:
This is fixing issue #1097

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,7 +2,7 @@
 #
 # Build with:
 # docker build --tag=kivy/buildozer .
-# 
+#
 # In order to give the container access to your current working directory
 # it must be mounted using the --volume option.
 # Run with (e.g. `buildozer --version`):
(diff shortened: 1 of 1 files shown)


AndreMiras:
Thanks for the report and PR. I'd like to see if I can reproduce it locally and if the changes fix it like we think.
Also if theses are really buildozer Python dependencies, then maybe it should be part of `setup.py`. I'd like to investigate a bit further before we merge

Sirfanas:
@AndreMiras I think it's not a buildozer requirements but a python-for-android requirements

AndreMiras:
Yes then this is probably were it should be handled. But buildozer relation with p4a is a bit tricky in that sense, still there's a way

Sirfanas:
@AndreMiras I think a good way to implement it would be to add everything in the p4a's requirements, cause it's used in it so it make sense
And to improve compilation time we can add them in the buildozer's Dockerfile, as it always need p4a
In this way when we manually install then every requirements would be installed, and Docker image's we'll be ready "out of the box", so it don't need any install to run command ("android clean" in this situation)

PR #1100, opened 2020-05-11
:rotating_light: Build integration test

AndreMiras:
Also fixes CHANGELOG.md descending PR order

### DIFF
--- a/.github/workflows/test_python.yml
+++ b/.github/workflows/test_python.yml
@@ -62,6 +62,10 @@ jobs:
       run: |
         sed "s/# android.accept_sdk_license = False/android.accept_sdk_license = True/" -i buildozer.spec
         buildozer android p4a -- --help
+    - name: buildozer android debug
+      run: |
+        touch main.py
+        buildozer android debug
 
   Python2:
     runs-on: ubuntu-latest
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 
 **Fixed bugs:**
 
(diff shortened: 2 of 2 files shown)


CHANGELOG.md
```diff
@@ -36,20 +37,24 @@
 
 **Merged pull requests:**
 
-- Android gradle build: fix apk name [\#1025](https://github.com/kivy/buildozer/pull/1025) ([SomberNight](https://github.com/SomberNight))
```
AndreMiras, line 39: Jus reordered from ascending to descending

.github/workflows/test_python.yml
```diff
@@ -62,6 +62,10 @@ jobs:
       run: |
         sed "s/# android.accept_sdk_license = False/android.accept_sdk_license = True/" -i buildozer.spec
         buildozer android p4a -- --help
+    - name: buildozer android debug
+      run: |
+        touch main.py
+        buildozer android debug
```
AndreMiras, line 68: Adds a 11 minutes overhead, but it's probably worth trying until we have proper unit test coverage :cry:

opacam: Oh boy, this is quite of time, but you are right, better this than nothing :+1:

AndreMiras:
Yes indeed it's quite some time and most of what we do here is already tested in p4a. But for the lack of having nice unit test coverage we integration test. That will give us confidence that at least the minimum is working.
Thank for the review! :rocket:

PR #1111, opened 2020-05-17
Update README.md

tshirtman:
don't advise using sudo pip!


AndreMiras:
Integration test got flaky with a:
```
OSError: [Errno 101] Network is unreachable
```
I think we can ignore and we will do workaround if it happens more often in the future

tshirtman:
yeah, i don't think that's caused by this PR, but sad :panda_face:

PR #1123, opened 2020-05-25
Update instalation.rst for Windows 10

arqeco:
Instructions for installing Buildozer on Windows 10

### DIFF
--- a/docs/source/installation.rst
+++ b/docs/source/installation.rst
@@ -27,6 +27,47 @@ Android on Ubuntu 18.04 (64bit)
     # add the following line at the end of your ~/.bashrc file
     export PATH=$PATH:~/.local/bin/
 
+Android on Windows 10
+~~~~~~~~~~~~~~~~~~~~~
+
+To use buildozer in Windows 10 you need first to enable Windows Subsystem for Linux (WSL) and install a Linux distribution: https://docs.microsoft.com/en-us/windows/wsl/install-win10.
+
+These instructions were tested with WSL 1 and Ubuntu 18.04 LTS. 
+
+With WSL and Ubuntu installed on your Windows 10 machine, open Ubuntu and run these commands:
+
+::
+
+    sudo apt update
+    sudo apt install -y git zip unzip openjdk-8-jdk python3-pip autoconf libtool pkg-config zlib1g-dev libncurses5-dev libncursesw5-dev libtinfo5 cmake libffi-dev libssl-dev
+    # Use here the python version you need
+    sudo apt install -y python3.7-venv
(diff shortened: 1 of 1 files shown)


docs/source/installation.rst
```diff
@@ -27,6 +27,47 @@ Android on Ubuntu 18.04 (64bit)
...
+    python3.7 -m venv venv-buildozer
+    source venv/bin/activate
+    python -m pip install --upgrade pip
+    python -m pip install --upgrade wheel
+    python -m pip install --upgrade cython 
+    python -m pip install --upgrade virtualenv 
+    python -m pip install --upgrade buildozer
+    # Add the following line at the end of your ~/.bashrc file
+    export PATH=$PATH:~/.local/bin/
+    # Restart your WSL terminal to enable the path change
```
AndreMiras, line 57: I feel like most of this could reuse what already holds true in the Ubuntu install section above, that way we don't repeat ourselves

docs/source/installation.rst
```diff
@@ -27,6 +27,47 @@ Android on Ubuntu 18.04 (64bit)
...
+
+Now you need to install the Windows version of ADB (Android Debug Bridge):
```
AndreMiras, line 59: isn't the version "shipped" by buildozer enough?

docs/source/installation.rst
arqeco: Yes

docs/source/installation.rst
arqeco: In my tests ADB for Windows was needed to detect the USB connected Android device. Maybe Windows Subsystem for Linux does not have direct access to hardware. ADB in Windows detects the USB device and start the daemon. ADB in buildozer (in WSL) connects to the daemon and access the USB device.

docs/source/installation.rst
arqeco: I 'm gonna change it

docs/source/installation.rst
tshirtman: yes, i did spend some time supporting an user in that case sometime ago, WSL indeed doesn't have USB support, so the only way is to use the adb daemon started in windows.

tshirtman: it's probably a good idea to add a note so they get the same adb version as the one installed by buildozer, even if it's probably going to happen by default most of the time, as if they do not, every time they will run one adb on any side, this one will kill the daemon and restart it, so it won't work.

arqeco:
Required changes were made

akshayaurora (approved):

akshayaurora:
@arqeco thanks 👍

PR #1119, opened 2020-05-21
Content providers

pavelsof:
This PR complements kivy/python-for-android#2200, adding two config options that can be used in `buildozer.spec`:

- `android.manifest.content_providers` expects a path to a file, the contents of which will be inserted into the `<application>` tag of the Android manifest.
- `android.add_xml_resources` expects a list of file paths to be copied into the `src/main/res/xml` subdirectory of the build, so that they can be referenced in the Android manifest.

As an example of how it would look like from the viewpoint of a developer using Kivy, one may take a look at pavelsof/mobile-wormhole@d0c100e1c77a73d3a2864a000ee6169fbaffdb69.

PR #1126, opened 2020-05-27
:sparkles: Adds p4a --numeric-version support

AndreMiras:
This flag is available in p4a and can be useful to have buildozer side too.
Also refs #1079

### DIFF
--- a/buildozer/default.spec
+++ b/buildozer/default.spec
@@ -221,6 +221,10 @@ fullscreen = 0
 # (str) The Android arch to build for, choices: armeabi-v7a, arm64-v8a, x86, x86_64
 android.arch = armeabi-v7a
 
+# (int) overrides automatic versionCode computation (used in build.gradle)
+# this is not the same as app version and should only be edited if you know what you're doing
+# android.numeric_version = 1
+
 #
 # Python for android (p4a) specific
 #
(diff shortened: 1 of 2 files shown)


AndreMiras:
I'll add the unit tests in another dedicated PR after addressing https://github.com/kivy/buildozer/pull/1119#issuecomment-634809317

buildozer/default.spec
```diff
@@ -221,6 +221,9 @@ fullscreen = 0
 # (str) The Android arch to build for, choices: armeabi-v7a, arm64-v8a, x86, x86_64
 android.arch = armeabi-v7a
 
+# (int) overrides automatic versionCode computation (used in build.gradle)
+# android.numeric_version = 1
```
inclement, line 225: Actually could you add a comment noting that this is not the same as the app version, and should only be edited if you know what you're doing?

AndreMiras: Good point I've added `this is not the same as app version and should only be edited if you know what're doing`

inclement: Typo "what're" -> "what you're" :p

AndreMiras: And this is why I love code review when it's possible
I've rushed it because I'm exited about the feature :)
Thanks addressed!

inclement: Now it's "yo're"

AndreMiras: Yes I saw it as I pushed and already fixed :facepalm:

AndreMiras: I'm happy to push force as many time as needed :smile:

PR #1129, opened 2020-05-29
:white_check_mark: Tests android.numeric_version config

AndreMiras:
Follow-up for #1126

### DIFF
--- a/tests/targets/test_android.py
+++ b/tests/targets/test_android.py
@@ -1,4 +1,5 @@
 import os
+import re
 import tempfile
 from unittest import mock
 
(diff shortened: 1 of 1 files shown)


tests/targets/test_android.py
```diff
@@ -93,11 +94,8 @@ def init_target(self, options=None):
         spec = []
         for line in default_spec:
             if line.strip():
-                key = line.split()[0]
-
-                if key.startswith('#'):
-                    key = key[1:]
-
+                match = re.search(r'[#\s]?([a-z_\.]+)', line)
+                key = match and match.group(1)
```
AndreMiras, line 98: More accurate (handle spaces just after the `#` comment), yet more concise version.
Works on matching:
```
#config.option = foo
```
As well as:
```
# config.option = foo
```
Both would make `key = "config.option"`

PR #1130, opened 2020-05-30
Modify android NDK version regex to match e.g. '21'

pnhofmann:


AndreMiras:
Thanks for the PR!
I've restarted the build as it seemed to failed for unrelated reason. Let's see how it goes.
Contribution tip: you can reference the task your are solving in your commit message (or body) e.g. adding "fixes #1130" at the end

PR #1131, opened 2020-05-30
:white_check_mark: Also integration test on macOS

AndreMiras:
Note the `sed` expression got updated to work for both OS.
Also workaround openssl lib install on macOS, the error was:
```
  RAN: /bin/bash -c 'venv/bin/pip install Cython'

  STDOUT:
WARNING: pip is configured with locations that require TLS/SSL, however the ssl module in Python is not available.
Collecting Cython
  WARNING: Retrying (Retry(total=4, connect=None, read=None, redirect=None, status=None)) after connection broken by 'SSLError("Can't connect to HTTPS URL because the SSL module is not available.")': /simple/cython/
  WARNING: Retrying (Retry(total=3, connect=None, read=None, redirect=None, status=None)) after connection broken by 'SSLError("Can't connect to HTTPS URL because the SSL module is not available.")': /simple/cython/
  WARNING: Retrying (Retry(total=2, connect=None, read=None, redirect=None, status=None)) after connection broken by 'SSLError("Can't connect to HTTPS URL because the SSL module is not available.")': /simple/cython/
  WARNING: Retrying (Retry(total=1, connect=None, read=None, redirect=None, status=None)) after connection broken by 'SSLError("Can't connect to HTTPS URL because the SSL module is not available.")': /simple/cython/
  WARNING: Retrying (Retry(total=0, connect=None, read=None, redirect=None, status=None)) after connection broken by 'SSLError("Can't connect to HTTPS URL because the SSL module is not available.")': /simple/cython/
  Could not fetch URL https://pypi.org/simple/cython/: There was a problem confirming the ssl certificate: HTTPSConnectionPool(host='pypi.org', port=443): Max retries exceeded with url: /simple/cython/ (Caused by SSLError("Can't connect to HTTPS URL because the SSL module is not available.")) - skipping
  ERROR: Could not find a version that satisfies the requirement Cython (from versions: none)
ERROR: No matching distribution found for Cython
WARNING: pip is configured with locations that require TLS/SSL, however the ssl module in Python is not available.
Could not fetch URL https://pypi.org/simple/pip/: There was a problem confirming the ssl certificate: HTTPSConnectionPool(host='pypi.org', port=443): Max retries exceeded with url: /simple/pip/ (Caused by SSLError("Can't connect to HTTPS URL because the SSL module is not available.")) - skipping
```

### DIFF
--- a/.github/workflows/test_python.yml
+++ b/.github/workflows/test_python.yml
@@ -45,7 +45,12 @@ jobs:
       run: docker run kivy/buildozer --version
 
   Integration:
-    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        os:
+          - 'ubuntu-latest'
+          - 'macOs-latest'
+    runs-on: ${{ matrix.os }}
     steps:
     - name: Setup python
       uses: actions/setup-python@v2
@@ -60,8 +65,17 @@ jobs:
     - run: buildozer init
     - name: SDK, NDK and p4a download
       run: |
-        sed "s/# android.accept_sdk_license = False/android.accept_sdk_license = True/" -i buildozer.spec
+        sed -i.bak "s/# android.accept_sdk_license = False/android.accept_sdk_license = True/" buildozer.spec
         buildozer android p4a -- --help
+    # Install OS specific dependencies
+    - name: Install Linux dependencies
+      if: matrix.os == 'ubuntu-latest'
+      run: sudo apt -y install automake
+    - name: Install macOS dependencies
+      if: matrix.os == 'macOS-latest'
+      run: |
+        brew install automake
+        sudo ln -sfn /usr/local/opt/openssl /usr/local/ssl
     - name: buildozer android debug
       run: |
         touch main.py
--- a/docs/source/installation.rst
+++ b/docs/source/installation.rst
@@ -13,7 +13,7 @@ First, install the buildozer project with::
 Targeting Android
 -----------------
 
-Android on Ubuntu 18.04 (64bit)
+Android on Ubuntu 20.04 (64bit)
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 (expected to work as well in later version, but only regularly tested in the latest LTS)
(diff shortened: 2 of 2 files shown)


.github/workflows/test_python.yml
```diff
@@ -60,8 +65,17 @@ jobs:
...
         buildozer android p4a -- --help
+    # Install OS specific dependencies
+    - name: Install Linux dependencies
+      if: matrix.os == 'ubuntu-latest'
+      run: sudo apt -y install automake
+    - name: Install macOS dependencies
+      if: matrix.os == 'macOS-latest'
+      run: |
+        brew install automake
+        sudo ln -sfn /usr/local/opt/openssl /usr/local/ssl
```
AndreMiras, line 78: This is a workaround that could easily be integrated p4a side by patching Python to add `/usr/local/opt/openssl` to the search path.
I actually think it should be a CPython upstream fix, I may look into this in the future

.github/workflows/test_python.yml
```diff
@@ -60,8 +65,17 @@ jobs:
     - run: buildozer init
     - name: SDK, NDK and p4a download
       run: |
-        sed "s/# android.accept_sdk_license = False/android.accept_sdk_license = True/" -i buildozer.spec
+        sed -i.bak "s/# android.accept_sdk_license = False/android.accept_sdk_license = True/" buildozer.spec
```
AndreMiras, line 68: sed in macOS doesn't work the same way as the one in Linux, the `-i.bak` is doing the trick to work on both

.github/workflows/test_python.yml
inclement: That's interesting indeed. Did macOS move this recently? Anyway, seems like a good fix for CI purposes.

Is this something users are going to have problems with?

AndreMiras: I assumed user may have problem with it, even though I didn't directly see the problem yet on my macOS.
What I saw is without this fix the python3 `config.log` clearly show it didn't find the openssl lib. However later I didn't get the pip SSL error. However on the CI I do get the error :thinking: 
So yes I may pull request this CPython upstream because the fix is easy

PR #1135, opened 2020-05-31
:arrow_up: Migrates to Command line tools

AndreMiras:
The SDK Tools package is deprecated and no longer receiving updates:
https://developer.android.com/studio/releases/sdk-tools
Also drops the "tools" package install as it's considered obsolete.

Migrates to command line tools:
https://developer.android.com/studio#cmdline-tools
https://developer.android.com/studio/command-line

Makes it possible to use both OpenJDK 8 and more recent versions.
Tested successfully in Arch with OpenJDK 10.

### DIFF
--- a/buildozer/targets/android.py
+++ b/buildozer/targets/android.py
@@ -44,7 +44,7 @@ DEPRECATED_TOKENS = (('app', 'android.sdk'), )
 # because it doesn't seem to matter much, it is normally correct to
 # download once then update all the components as buildozer already
 # does.
-DEFAULT_SDK_TAG = '4333796'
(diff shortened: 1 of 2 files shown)


inclement:
Awesome. Does that mean p4a worked with jdk 10 all along?

AndreMiras:
Thanks for the review.
Yes totally, I could compile and run an app. I didn't get as greedy as trying with even more recent JDK versions, but I would not be surprised if it worked.
The idea is to see how this is going for a while keeping CI and documentation on OpenJDK 8, then eventually update both

PR #1142, opened 2020-06-06
:fire: Drops garden support

AndreMiras:
Garden are now installed like normal packages, refs:
https://kivy.org/doc/stable/api-kivy.garden.html#update-to-garden-structure

### DIFF
--- a/buildozer/__init__.py
+++ b/buildozer/__init__.py
@@ -14,6 +14,7 @@ import sys
 import select
 import codecs
 import textwrap
+import warnings
 from buildozer.jsonstore import JsonStore
 from sys import stdout, stderr, exit
 from re import search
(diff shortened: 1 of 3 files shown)


AndreMiras:
I kept it simple and only ditched the feature without further notice. We saw too many users being confused recently with the way to install/use gardens. Furthermore most used gardens are already migrated to the new structure

buildozer/__init__.py
```diff
@@ -508,46 +505,6 @@ def _install_application_requirement(self, module):
                  env=self.env_venv,
                  cwd=self.buildozer_dir)
 
-    def check_garden_requirements(self):
-        '''Ensure required garden packages are available to be included.
-        '''
-        garden_requirements = self.config.getlist('app',
-                'garden_requirements', '')
-
```
opacam, line 516: Maybe we should emit some kind of `removed warning` (or maybe raise an exception) in case that we found `garden_requirements` inside a `buildozer.spec` file. This way, users could move the garden requirements...otherwise all garden packages will be silently omitted, am I right?

AndreMiras: Addressed (but untested)

PR #1154, opened 2020-06-14
Move Buildozer into own file.

Julian-O:
__init__.py files are responsible for:

a) Making sure clients don't need to understand the substructure of the package.
b) Determining which parts of the module should be shared with the client.

Due to the reg-exp based automated version detection, this one also keeps track of the version of the module.

The rest of the concerns of the module have been separated into a separate file.


Julian-O:
Ugh, different version number systems conflicting.

PR #1156, opened 2020-06-15
Cleanup of imports

Julian-O:
PEP8 states:

Imports are always put at the top of the file, just after any module comments and docstrings, and before module globals and constants.

Imports should be grouped in the following order:
1. Standard library imports.
2. Related third party imports.
3. Local application/library specific imports.
You should put a blank line between each group of imports.

---

Followed this convention.
Additionally sorted alphabetically, to help find cases of double imports.

Where there were double imports:
- didn't go as far as removing them, but ensured all references to explicitly-named imports were using them.
- removed `import sys` (in remote.py)

android.py imported platform.executable which was then hidden by a local with the same identifier. Renamed the local to avoid aliasing.

### DIFF
--- a/buildozer/targets/android.py
+++ b/buildozer/targets/android.py
@@ -6,7 +6,24 @@ import sys
 if sys.platform == 'win32':
     raise NotImplementedError('Windows platform not yet working for Android')
 
-from platform import uname
+import ast
+from distutils.version import LooseVersion
+import io
+from glob import glob
+from pipes import quote
+import os
+from os import environ
+from os.path import exists, join, realpath, expanduser, basename, relpath
+from platform import architecture, uname
+import re
+from shutil import copyfile, rmtree
+from sys import platform, executable
+import traceback
+
+from buildozer import BuildozerException, USE_COLOR
+from buildozer.target import Target
+from buildozer.libs.version import parse
+
 WSL = 'Microsoft' in uname()[2]
 
 ANDROID_API = '27'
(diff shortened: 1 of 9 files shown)


Julian-O:
First I've heard of it. Researching it now.

Julian-O:
Ok, it is much stricter - it sorts `from package import symbol` separately from `import package`.
It also sorts the symbols on the line. 

Also, it'll need to be added to the test_requires in the setup file.

Working on it.

AndreMiras:
Yes it is more strict, even though it can be tuned. Actually let's wait before you spend further time on it, it's a contentious topic :smile: 
My position is:
I'd like to respect the PEP8 as much as possible.
But for me there's no point to make code changes back and forth if we don't have a linter to demonstrate compliance. Last time I cross checked with other developers, they were not so eager to make this move. The import mess we have has also been bugging me for a long time. But I don't feel like doing the effort of cleaning it up if there's no tool preventing more mess in the future, that's why I never took action in the end.
So let's discuss with other devs before we introduce the change.
Mainly I'd like to have the feedback from @inclement and/or @tito on this matter

Julian-O:
Okay, but the progress I've made so far confirms that isort can act as a linter.
I've made it a separate run in tox. The dependency appears there, not in setup.py as I suggested earlier.

akshayaurora:
@Julian-O would you please rebase from current master.


@AndreMiras I would vote +1 for the change introducing isort.

akshayaurora:
@Julian-O would you please rebase from current master.


@AndreMiras I would vote +1 for the change introducing isort.

buildozer/targets/android.py
```diff
@@ -70,8 +69,8 @@ def __init__(self, *args, **kwargs):
             'app', 'android.arch', DEFAULT_ARCH)
         self._build_dir = join(
             self.buildozer.platform_dir, 'build-{}'.format(self._arch))
-        executable = sys.executable or 'python'
-        self._p4a_cmd = '{} -m pythonforandroid.toolchain '.format(executable)
+        _executable = executable or 'python'
```
inclement, line 72: As a general comment, this is exactly why importing every name explicitly isn't necessarily the best idea - using up `executable` in the module namespace is fairly pointless, there's nothing wrong with referencing `sys.executable`. More generally, I think importing explicitly from `os`, `sys` etc. provides no benefit and if anything makes the code less clear.

PR #1160, opened 2020-06-20
:green_apple: Improves iOS support

AndreMiras:
Updates & fixes support for iOS target, covers the following:
- Build integration test
- Updates to new kivy-ios structure
- Bumps to ios-deploy 1.10.0 (`make` replaced by `xcodebuild`)
- Makes it possible to toggle code signing (now disabled by default)

Note this is the first iteration of a longer serie.
Subsequent pull requests will try to improve the following:
- code signing process (toggle not fully integrated)
- actual deployment (not yet tested)
- unit tests (mainly `buildozer/targets/ios.py`)
- other forms of technical debt

### DIFF
--- a/.github/workflows/android.yml
+++ b/.github/workflows/android.yml
@@ -17,7 +17,7 @@ jobs:
     - name: Setup environment
       run: |
         pip install -e .
-        pip install Cython==0.29.19
+        pip install Cython
     - run: buildozer --help
     - run: buildozer init
     - name: SDK, NDK and p4a download
--- a/buildozer/default.spec
+++ b/buildozer/default.spec
@@ -22,7 +22,7 @@ source.include_exts = py,png,jpg,kv,atlas
 #source.exclude_exts = spec
 
 # (list) List of directory to exclude (let empty to not exclude anything)
-#source.exclude_dirs = tests, bin
+#source.exclude_dirs = tests, bin, venv
 
 # (list) List of exclusions using pattern matching
 #source.exclude_patterns = license,images/*/*.jpg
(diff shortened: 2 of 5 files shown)


.github/workflows/android.yml
```diff
@@ -17,7 +17,7 @@ jobs:
     - name: Setup environment
       run: |
         pip install -e .
-        pip install Cython==0.29.19
+        pip install Cython
```
AndreMiras, line 20: We use to have issues years ago with Cython versions, but it's no longer the case, so I don't see a reason to pin it until issues are back

buildozer/default.spec
```diff
@@ -22,7 +22,7 @@ source.include_exts = py,png,jpg,kv,atlas
 #source.exclude_exts = spec
 
 # (list) List of directory to exclude (let empty to not exclude anything)
-#source.exclude_dirs = tests, bin
+#source.exclude_dirs = tests, bin, venv
```
AndreMiras, line 25: I too often see people doing reports because buildozer picked up some `venv` garbage, so here're some pointers for people who care to read through the default example :smile:

buildozer/default.spec
```diff
@@ -263,7 +263,10 @@ ios.kivy_ios_branch = master
 #ios.ios_deploy_dir = ../ios_deploy
 # Or specify URL and branch
 ios.ios_deploy_url = https://github.com/phonegap/ios-deploy
-ios.ios_deploy_branch = 1.7.0
+ios.ios_deploy_branch = 1.10.0
+
+# (bool) Whether or not to sign the code
+ios.codesign.allowed = false
```
AndreMiras, line 269: Two things here. First, I'm ditching the old C-style `0` or `1` fashion and I plan to do so through the entire code base in the future. The good thing is it's retro compatible, so people it won't change in behavior for people having `0` or `1` in their settings https://docs.python.org/3/library/configparser.html#supported-datatypes
Second I set this up to `false` by default while introducing the setting. This is a change in behavior, but will lead to less complains/questions from our users. This is because most of the time people just want to try things locally and don't need to sign and publish to the Apple store. As for the ones who went through all the struggle and actually made it to be 99% ready to publish to the store, they're usually good enough to figure out they need to change this setting to sign their app

buildozer/targets/ios.py
```diff
@@ -99,10 +102,21 @@ def install_platform(self):
                                                           branch='1.7.0',
```
AndreMiras, line 102: I decided to not address this DRY now, but it's on my radars for next pull requests

buildozer/default.spec
tito: I honestly though that codesign would be used in release only, never in debug mode

PR #1168, opened 2020-06-27
:white_check_mark: Unit test ios target

AndreMiras:
- basic ios target test
- refactors existing android test for code sharing
- changes `call_build_package()` to simple function

Grows `buildozer/targets/ios.py` coverage from 11% to 24%.
This setups the canvas for more tests to come later.

### DIFF
--- a/buildozer/targets/ios.py
+++ b/buildozer/targets/ios.py
@@ -3,9 +3,6 @@ iOS target, based on kivy-ios project
 '''
 
 import sys
-if sys.platform != 'darwin':
-    raise NotImplementedError('Windows platform not yet working for Android')
-
 import plistlib
 from buildozer import BuildozerCommandException
 from buildozer.target import Target, no_config
@@ -64,6 +61,8 @@ class TargetIos(Target):
     targetname = "ios"
 
     def check_requirements(self):
+        if sys.platform != "darwin":
+            raise NotImplementedError("Only macOS is supported for iOS target")
         checkbin = self.buildozer.checkbin
         cmd = self.buildozer.cmd
         executable = sys.executable or 'python'
--- a/tests/targets/test_android.py
+++ b/tests/targets/test_android.py
@@ -1,17 +1,15 @@
 import os
-import re
 import tempfile
 from unittest import mock
 
 import pytest
 
-import buildozer as buildozer_module
-from buildozer import Buildozer
 from buildozer.targets.android import TargetAndroid
-
-
-def patch_buildozer(method):
-    return mock.patch("buildozer.Buildozer.{method}".format(method=method))
+from tests.targets.utils import (
+    init_buildozer,
+    patch_buildozer,
+    patch_buildozer_checkbin,
+)
 
 
 def patch_buildozer_cmd():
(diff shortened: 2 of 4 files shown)


buildozer/targets/ios.py
```diff
@@ -3,9 +3,6 @@
 '''
 
 import sys
-if sys.platform != 'darwin':
-    raise NotImplementedError('Windows platform not yet working for Android')
```
AndreMiras, line 7: We got the message a bit wrong, but also this got moved to `check_requirements()` instead. It feels more appropriated and improves testability

tests/targets/test_android.py
```diff
@@ -52,11 +46,57 @@ def patch_platform(platform):
...
-class TestTargetAndroid:
+def init_target(temp_dir, options=None):
+    buildozer = init_buildozer(temp_dir, "android", options)
+    return TargetAndroid(buildozer)
 
-    @staticmethod
-    def default_specfile_path():
-        return os.path.join(os.path.dirname(buildozer_module.__file__), "default.spec")
+
+def call_build_package(target_android):
```
AndreMiras, line 54: Changed to a more "functional approach" to make it clear it only requires target_android and won't mutate self

tests/targets/test_android.py
```diff
@@ -52,11 +46,57 @@ def patch_platform(platform):
...
+        '_generate_whitelist'
+    ) as m_generate_whitelist, mock.patch(
+        'buildozer.targets.android.TargetAndroid.execute_build_package'
+    ) as m_execute_build_package, mock.patch(
+        'buildozer.targets.android.copyfile'
+    ) as m_copyfile, mock.patch(
+        'buildozer.targets.android.os.listdir'
+    ) as m_listdir:
+        m_listdir.return_value = ['30.0.0-rc2']
+        target_android.build_package()
```
inclement, line 81: I really dislike this with statement formatting :<

AndreMiras: sure, let me know the way you like it so I can change it to it

AndreMiras: I've updated it to the version with the backslash https://github.com/psf/black/blob/7403d95/docs/the_black_code_style.md#how-black-wraps-lines let me know what you think

tests/targets/test_android.py
```diff
@@ -52,11 +46,52 @@ def patch_platform(platform):
...
+    instance so that tests can easily check which command-line arguments
+    would be passed on to python-for-android's toolchain.
+    """
+    buildozer = target_android.buildozer
+    expected_dist_dir = (
+        '{buildozer_dir}/android/platform/build-armeabi-v7a/dists/myapp__armeabi-v7a'.format(
+        buildozer_dir=buildozer.buildozer_dir)
+    )
+
+    with \
```
inclement, line 69: Breaking the first line here seems a bit pointless, it actually leads to more indentation rather than less!

AndreMiras: I think it's because indentation needs to be a multiple a 4 and not under/over indented (unless the flake8 rule is commented which is the case) so if you want them aligned that's how you can address it.
I've changed it anyway, let me know what do you think

PR #1169, opened 2020-06-27
 added android.activity_class_name option

vesellov:
to make possible to use custom PythonActivity Java class

### DIFF
--- a/buildozer/__init__.py
+++ b/buildozer/__init__.py
@@ -309,7 +309,10 @@ class Buildozer:
                 if get_stdout:
                     ret_stdout.append(chunk)
                 if show_output:
-                    stdout.write(chunk.decode('utf-8', 'replace'))
+                    try:
+                        stdout.write(chunk.decode('utf-8', 'replace'))
+                    except UnicodeEncodeError:
+                        stdout.write('?' * len(chunk))
             if fd_stderr in readx:
                 chunk = process.stderr.read()
                 if not chunk:
@@ -317,7 +320,10 @@ class Buildozer:
                 if get_stderr:
                     ret_stderr.append(chunk)
                 if show_output:
-                    stderr.write(chunk.decode('utf-8', 'replace'))
+                    try:
+                        stderr.write(chunk.decode('utf-8', 'replace'))
+                    except UnicodeEncodeError:
+                        stderr.write('?' * len(chunk))
 
             stdout.flush()
             stderr.flush()
--- a/buildozer/default.spec
+++ b/buildozer/default.spec
@@ -192,6 +192,14 @@ fullscreen = 0
 # (str) launchMode to set for the main activity
 #android.manifest.launch_mode = standard
 
+# (str) Add a Network Security Configuration file path to AndroidManifest.xml
+# File with network configs must already exist in:
+# ./pythonforandroid/bootstraps/sdl2/build/src/main/res/xml/network_security_config.xml
+#android.manifest.network_security_config = "@xml/network_security_config"
+
+# (str) Indicate that app intends to use cleartext network traffic in AndroidManifest.xml
+#android.manifest.uses_cleartext_traffic = False
+
 # (list) Android additional libraries to copy into libs/armeabi
 #android.add_libs_armeabi = libs/android/*.so
 #android.add_libs_armeabi_v7a = libs/android-v7/*.so
--- a/buildozer/default.spec
+++ b/buildozer/default.spec
@@ -123,10 +123,14 @@ fullscreen = 0
 # android.accept_sdk_license = False
 
 # (str) Android entry point, default is ok for Kivy-based app
-#android.entrypoint = org.renpy.android.PythonActivity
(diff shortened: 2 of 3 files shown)


buildozer/targets/android.py
```diff
@@ -92,6 +92,16 @@ def __init__(self, *args, **kwargs):
         if port is not None:
             self.extra_p4a_args += ' --port={}'.format(port)
 
+        activity_package = self.buildozer.config.getdefault(
+            'app', 'android.activity_package', 'org.kivy.android')
```
AndreMiras, line 96: Should we also document both params by updating the `default.spec`?
The idea is to comment out an example there (I would use the default values)

buildozer/__init__.py
```diff
@@ -309,15 +309,21 @@ def cmd(self, command, **kwargs):
...
                 if not chunk:
                     break
                 if get_stderr:
                     ret_stderr.append(chunk)
                 if show_output:
-                    stderr.write(chunk.decode('utf-8', 'replace'))
+                    try:
+                        stderr.write(chunk.decode('utf-8', 'replace'))
+                    except UnicodeEncodeError:
+                        stderr.write('?' * len(chunk))
```
AndreMiras, line 326: I have no idea why we're doing this, could you elaborate? Or maybe write a test case for it?

buildozer/targets/android.py
vesellov: yes, i will do that later, @AndreMiras  👍

buildozer/__init__.py
vesellov: I got that error on my Ubuntu ...    It seems to me that in some cases sub-process that is executed by `buildozer` to run compile/build steps is returning non-ascii characters to `stdout` (and could be potentially to `stderr` as well) .   I do not remember which of the packages it was failing but that fix made it working.  probably not nicest solution, but at least it is a safe operation.

buildozer/targets/android.py
vesellov: updated the PR... but now i see there is some error in the build: `Aidl not found, please install it.`   I feel pretty confident that my changes have nothing to do with that...

AndreMiras: Indeed, that could be a temporary glitch, like network hiccup. I'll keep an eye on the CI and restart it. Thanks for pointing it out

buildozer/default.spec
```diff
@@ -114,19 +114,23 @@ fullscreen = 0
 # (bool) If True, then skip trying to update the Android sdk
 # This can be useful to avoid excess Internet downloads or save time
 # when an update is due and you just want to test/build your package
-# android.skip_update = False
+#android.skip_update = False
```
AndreMiras, line 117: don't touch theses lines since it's out of the context of the PR. And actually that's the reason the CI is failing:
```
The following packages can not be installed since their licenses or those of the packages they depend on were not accepted:
```

buildozer/default.spec
```diff
@@ -114,19 +114,23 @@ fullscreen = 0
...
 # (bool) If True, then automatically accept SDK license
 # agreements. This is intended for automation only. If set to False,
 # the default, you will be shown the license when first running
 # buildozer.
-# android.accept_sdk_license = False
+#android.accept_sdk_license = False
 
 # (str) Android entry point, default is ok for Kivy-based app
-#android.entrypoint = org.renpy.android.PythonActivity
+#android.entrypoint = org.kivy.android.PythonActivity
```
AndreMiras, line 126: Good catch, thanks!

buildozer/targets/android.py
AndreMiras: Restarted the CI and still failing, it's actually because of this change I think:
https://github.com/kivy/buildozer/pull/1169/files#r446697206
Yes this is fishy that such a change break it, but it's because that line is `sed`-ed I think :sweat_smile:

vesellov: oh....
i thought it was just a mistake and removed that extra space ...

buildozer/default.spec
vesellov: ok. put it back.. very strange

buildozer/targets/android.py
AndreMiras: It's definitely some inconsistencies I agree, but then some CI tweaks got built upon it :facepalm: so here we are :smile:

AndreMiras:
The iOS build glitched with:
```
[DEBUG   ] /Applications/Xcode_11.5.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/usr/include/tk.h:86:11: fatal error: 
[DEBUG   ]       'X11/Xlib.h' file not found
[DEBUG   ] #       include <X11/Xlib.h>
[DEBUG   ]                 ^~~~~~~~~~~~
[DEBUG   ] 1 error generated.
```
This is documented in https://github.com/actions/virtual-environments/issues/1088
I'll restart the build just now

vesellov:
i will rework it a bit more and also require a bit more changes in p4a first

Issue #881, opened 2019-04-16
How to pass "allow_backup" as parameter to python-for-android?


edsonitk:
Hi,

I need to avoid the automatic backup of android and I believe that if I change the allow_backup parameter in build.py it should work

Setup
Python: 2.7
Android
Buildozer:

AndreMiras:
Linking the p4a PR for reference https://github.com/kivy/python-for-android/pull/1438

AndreMiras:
Closed via #1206

PR #1206, opened 2020-08-10
Fix issue #881: Add android.allow_backup parameter to buildozer.spec

Jorilx:
Just added support for allow_backup inside buildozer.spec (issue #881)


tshirtman:
Thanks! Test appreciated!

PR #1223, opened 2020-09-08
separate build per android.arch

fritzz90:
> This prevent using the same build (armv7a) when building another arch using profile (arm64-v8a)


tito:
My approach was using templates in buildozer.spec::

[app@arm64-v8a]
android.arch = arm64-v8a
build_dir = .buildozer/build/arm64-v8a/

Then invocation::

buildozer android release
buildozer android release --profile arm64-v8a

tito (approved):

tito:
It is already on master.

PR #1225, opened 2020-09-11
Add option for controlling p4a distutils support, fixes #1224

### LINKED ISSUE #1224: Cannot tell p4a to use/not to use distutils integration
### Versions

* Python: 3.7
* OS: Ubuntu 20.04 
* Buildozer: git master (and pypi 1.2.0)

### Description

Cannot tell p4a to use distutils support (https://python-for-android.readthedocs.io/en/latest/distutils/) since there
is no way to pass the argument.

Also in the future p4a is going to default to enabling distutils support (now defaults to disabled) so in the future
things will behave differently unless the corresponding option is explicitly passed to p4a

rambo:
as per #1224 

  1. We need to pass the option explicitly to keep old default behaviour in the future
  2. user should be able to control whether distutils support is enabled or not.

### DIFF
--- a/buildozer/default.spec
+++ b/buildozer/default.spec
@@ -256,6 +256,13 @@ android.allow_backup = True
 # (int) port number to specify an explicit --port= p4a argument (eg for bootstrap flask)
 #p4a.port =
 
+# Control passing the --use-setup-py vs --ignore-setup-py to p4a
+# "in the future" --use-setup-py is going to be the default behaviour in p4a, right now it is not
+# Setting this to false will pass --ignore-setup-py, true will pass --use-setup-py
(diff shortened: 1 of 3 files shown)


AndreMiras (approved):
LGTM, thanks for the pull request :pray:

AndreMiras (changes requested):
Oops I overlooked one thing. Do you mind fixing the broken test `tests/targets/test_android.py`

rambo:
Yeah noticed that too from the CI report, I'm having weird problems with tox in the Ubuntu VM I'm doing this on so forgot to run tests (and in python-for-android -side just running py.test also throws up in weird ways so I didn't even remember to try that here)

AndreMiras (approved):
LGTM, thanks!

PR #1238, opened 2020-10-15
Add android.adb_args option

germn:
Now, if a buildozer is in a Docker container, to make it work with a Linux host you [have to pass several unobvious flags to the Docker](https://github.com/tshirtman/Buildozer-docker#installing). I wasn't able to make it work with Windows host at all.

However, it's possible to achieve the same much easier, if we could [pass a flag to the adb command](https://stackoverflow.com/a/53765498/1113207). It makes buildozer in a Docker work without any additional Docker flags and on any host (AFAIK).

The PR adds a possibility to pass an argument(s) to the adb. Uncomment the default value (`-H host.docker.internal`) and run `buildozer android deploy` from inside Docker to see the effect.

### DIFF
--- a/buildozer/default.spec
+++ b/buildozer/default.spec
@@ -219,6 +219,9 @@ fullscreen = 0
 # (str) Android logcat filters to use
 #android.logcat_filters = *:S python:D
 
+# (str) Android additional adb arguments
+#android.adb_args = -H host.docker.internal
+
 # (bool) Copy library instead of making a libpymodules.so
 #android.copy_libs = 1
 
--- a/buildozer/targets/android.py
+++ b/buildozer/targets/android.py
@@ -262,6 +262,12 @@ class TargetAndroid(Target):
                     'zlib headers must be installed, '
                     'run: sudo apt-get install zlib1g-dev')
 
+        # Adb arguments:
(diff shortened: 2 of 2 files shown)


broccoliboy:
I would love to see this PR merged. I was just struggling with the problem of deploying to an android device from inside a [VS Code devcontainer](https://code.visualstudio.com/docs/remote/containers) based on the [official kivy/buildozer container](https://hub.docker.com/r/kivy/buildozer). With this PR I can configure buildozer in the devcontainer to use the host's adb server which has already been configured with my devices. This enables seamless build, deploy, and logcat viewing from within the devcontainer.

adin234:
any updates on this one?

germn:
> any updates on this one?

Let's try to ping few people who can merge: @inclement @AndreMiras @tito 

Note, however, that instead of waiting you can create your fork of buildozer, merge the PR and use it now.
Or you can, for example, [install buildozer from the PR's branch](https://stackoverflow.com/a/20101940/1113207):
`pip install https://github.com/germn/buildozer/archive/add_adb_args_option.zip`

AndreMiras (approved):
LGTM thanks for the proposal

PR #1263, opened 2020-12-17
Revert "Add manifestPlaceholders feature"

inclement:
This reverts commit b60055cdc073238af66552164775598a84daa3a6.

This commit was making buildozer crash with anything except p4a master.

The new feature is fine, but we should presumably not pass any argument unless the user specifically adds it. Buildozer could also check that the p4a version is compatible with this argument.

I strongly think we should revert immediately and fix the change later, as this is making buildozer fail for people.

### DIFF
--- a/buildozer/default.spec
+++ b/buildozer/default.spec
@@ -235,12 +235,6 @@ android.allow_backup = True
 # (str) XML file for custom backup rules (see official auto backup documentation)
 # android.backup_rules =
 
-# (str) If you need to insert variables into your AndroidManifest.xml file,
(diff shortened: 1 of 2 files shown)


inclement:
@AndreMiras Any objection?

Also FAO @misl6

AndreMiras (approved):
Thanks for the heads up :+1:

misl6:
Nothing against it, but i think #1265 it's just better.

This avoid adding `--manifest-placeholders` unless specified by the user.

PR #1269, opened 2020-12-24
Let buildozer.spec files pin to a specific p4a commit hash

xloem:
This PR adds a configuration option `p4a.commit` that can be used to require the same exact python-for-android revision for every build.  (edit: rebased to remove spurious commits)


akshayaurora (approved):

akshayaurora:
@xloem 👍

PR #1272, opened 2021-01-04
Allow displaying only the logcat of our app.

tshirtman:
Todo
- [ ] test with a service
- [ ] consider making default
- [ ] test for robustness

To discuss
- [ ] should it exit automatically like it does currently? what if user reruns manually?
- [ ] consider the cost of additional complexity vs usage simplification for users

### DIFF
--- a/buildozer/__init__.py
+++ b/buildozer/__init__.py
@@ -19,7 +19,7 @@ from buildozer.jsonstore import JsonStore
 from sys import stdout, stderr, exit
 from re import search
 from os.path import join, exists, dirname, realpath, splitext, expanduser
-from subprocess import Popen, PIPE
+from subprocess import Popen, PIPE, TimeoutExpired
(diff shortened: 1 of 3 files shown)


AndreMiras (dismissed):
I don't see why not, except for your last point code complexity vs benefit.
I can already see some benefit as inexperienced developers often find it hard to look the adb logs so that could potentially help.
But this feature could be redundant with the built-in adb filters, while the implementation is totally different, the result might look the same.
Still nothing against that PR

tshirtman:
I didn't try, but i think this should have one advantage over filtering, as it should also display lower level errors, like stacktraces, that can be hidden when using the "Python" prefix as a filter, also, in the hopefuly plausible of the users having multiple python applications installed, it should filter only the one they are trying to debug.

AndreMiras (approved):
LGTM

vesellov (approved):

PR #1306, opened 2021-03-16
Update iOS code signing

syrykh:
During testing of IPA build & sign couple of points for improvement where identified.

1. Commit "Add iOS options for manual code signing" adds some flexibility for automated builds in specifying code signing options like the development team and the provisioning profile to use for signing. The change is backward-compatible with old specs: if those parameters are not set, they are omitted from xcodebuild command invocation.
2. Commit "Update xcodebuild exportArchive options" removes exportFormat parameter of the exportArchive, which was [removed in Xcode 8.3](https://developer.apple.com/library/archive/releasenotes/DeveloperTools/RN-Xcode/Chapters/Introduction.html#//apple_ref/doc/uid/TP40001051-CH1-SW162) and adds exportOptionsPlist parameter instead.
3. Commit "iOS OTA manifest generation for in-house distribution" enables generation of manifest.plist to be used for in-house OTA ipa distribution.

### DIFF
--- a/buildozer/targets/ios.py
+++ b/buildozer/targets/ios.py
@@ -274,14 +274,13 @@ class TargetIos(Target):
             cwd=build_dir)
 
         self.buildozer.info('Creating IPA...')
-        self.xcodebuild((
-                ' -exportArchive'
-                ' -exportFormat IPA'
-                ' -archivePath "{xcarchive}"'
-                ' -exportPath "{ipa}"'
-                ' CODE_SIGN_IDENTITY={ioscodesign}'
-                ' ENABLE_BITCODE=NO'
-            ).format(xcarchive=xcarchive, ipa=ipa_tmp, ioscodesign=ioscodesign),
+        self.xcodebuild(
+            '-exportArchive',
+            f'-archivePath "{xcarchive}"',
+            f'-exportOptionsPlist "{plist_rfn}"',
+            f'-exportPath "{ipa_tmp}"',
+            f'CODE_SIGN_IDENTITY={ioscodesign}'
+            'ENABLE_BITCODE=NO',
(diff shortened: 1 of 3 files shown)


tests/targets/test_ios.py
```diff
@@ -196,13 +195,6 @@ def test_build_package_no_signature(self):
                 'You must fill the "ios.codesign.debug" token.'
             )
         ]
-        assert m_xcodebuild.call_args_list == [
```
syrykh, line 199: Because xcodebuild function signature has changed, instead of writing ...
```python
assert m_xcodebuild.call_args_list == [
    mock.call(
        "-configuration Debug",
        "ENABLE_BITCODE=NO",
        "CODE_SIGNING_ALLOWED=NO",
        None,
        None,
        None,
        "clean build",
        cwd="/ios/dir/myapp-ios",
    )
]
```
... I've moved the expectations to the `m_cmd` mock.

buildozer/targets/ios.py
```diff
@@ -115,15 +115,30 @@ def toolchain(self, cmd, **kwargs):
         kwargs.setdefault('cwd', self.ios_dir)
         return self.buildozer.cmd(self._toolchain_cmd + cmd, **kwargs)
 
-    def xcodebuild(self, cmd='', **kwargs):
-        return self.buildozer.cmd(self._xcodebuild_cmd + cmd, **kwargs)
+    def xcodebuild(self, *args, **kwargs):
+        return self.buildozer.cmd(self._xcodebuild_cmd + ' '.join(arg for arg in args if arg is not None), **kwargs)
```
syrykh, line 119: This approach allows easily omit unspecified (None) arguments, which makes later invocation cleaner.

PR #1311, opened 2021-03-30
Add enable_androidx

RobertFlatt:
Complement to https://github.com/kivy/python-for-android/pull/2385 now merged into p4a develop.

### DIFF
--- a/buildozer/default.spec
+++ b/buildozer/default.spec
@@ -164,7 +164,7 @@ fullscreen = 0
 
 # (bool) Enable AndroidX support. Enable when 'android.gradle_dependencies'
 # contains an 'androidx' package, or any package from Kotlin source.
-# android.enable_androidx requires android.api >= 28 
+# android.enable_androidx requires android.api >= 28
 #android.enable_androidx = False
 
 # (list) add java compile options
--- a/buildozer/default.spec
+++ b/buildozer/default.spec
@@ -162,6 +162,11 @@ fullscreen = 0
 # (list) Gradle dependencies to add
 #android.gradle_dependencies =
 
+# (bool) Enable AndroidX support. Enable when 'android.gradle_dependencies'
+# contains an 'androidx' package, or any package from Kotlin source.
+# android.enable_androidx requires android.api >= 28 
+#android.enable_androidx = False
+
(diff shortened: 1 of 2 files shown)


buildozer/default.spec
```diff
@@ -162,6 +162,11 @@ fullscreen = 0
 # (list) Gradle dependencies to add
 #android.gradle_dependencies =
 
+# (bool) Enable AndroidX support. Enable when 'android.gradle_dependencies'
+# contains an 'androidx' package, or any package from Kotlin source.
+# android.enable_androidx requires android.api >= 28
```
tshirtman, line 167: maybe pep8 is unhappy about the trailing space
```suggestion
# android.enable_androidx requires android.api >= 28
```

buildozer/targets/android.py
```diff
@@ -1167,6 +1167,13 @@ def build_package(self):
         if wakelock:
             build_cmd += [("--wakelock", )]
 
+        # AndroidX ?
+        enable_androidx = config.getbooldefault('app',
```
tshirtman, line 1171: Pretty sure this is legal, but
```python
        enable_androidx = config.getbooldefault(
            'app', 'android.enable_androidx', False)
```
would be more consistent to the code around.

tshirtman (commented):
running `pep8` (or rather `pycodestyle`) localy on this module would help make sure what makes the CI unhappy, but i think it's worth trying the trailing space for a start.

buildozer/default.spec
RobertFlatt: Thank you. It was trailing whitespace.

allerter:
Shouldn't there be a check for the target API level? It could help users avoid unhelpful errors later in the build process.
```python
    if int(self.android_api) < 28:
        raise BuildozerException("android.enable_androidx requires android.api >= 28.")
```

RobertFlatt:
>Shouldn't there be a check for the target API level? 

I expect there _could_ be

The same issue came up with the p4a PR review  https://github.com/kivy/python-for-android/pull/2385#pullrequestreview-624168321 where the review comment was "I guess it doesn't really matter".

tshirtman (approved):

PR #1312, opened 2021-04-07
android: adaptive icon/launcher: expose p4a options in buildozer.spec

SomberNight:
see related p4a PR: https://github.com/kivy/python-for-android/pull/2446

This exposes the p4a options added in the referenced PR in `buildozer.spec`.

### DIFF
--- a/buildozer/default.spec
+++ b/buildozer/default.spec
@@ -87,6 +87,10 @@ fullscreen = 0
 # Lottie files can be created using various tools, like Adobe After Effect or Synfig.
 #android.presplash_lottie = "path/to/lottie/file.json"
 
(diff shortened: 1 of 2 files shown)


tshirtman (dismissed):
LGTM
sorry for late feedback :)
If you want to add the option to the default.spec, so it's easier for users to discover, but otherwise looks fine.

SomberNight:
> If you want to add the option to the default.spec, so it's easier for users to discover

I've added the option to `default.spec`.
Not sure if it would be better directly after the `icon.filename` option, or somewhere in the Android section;
I've put it into the Android section atm.

tshirtman:
yeah, i don’t know if there is something similar for IOS we need to support, but that should be fine for now, thanks!

tshirtman (approved):

PR #1317, opened 2021-04-20
iOS OTA manifest generation for in-house app distribution

syrykh:
This PR enables `manifest.plist` iOS generation, which can be used for installing the IPA via Safari.


Zen-CODE (approved):
Neat! That's a very useful feature...+1. 

Note: Cannot test that it actually works, but seems worth trying.

PR #1319, opened 2021-04-22
use HTTPS urls

obfusk:


obfusk:
@AndreMiras @tshirtman

Most of those URLs -- e.g. in the README -- aren't very important.

But downloading e.g. `ant` and `kivy` over unencrypted http w/o checking e.g. a SHA256 checksum is pretty bad security practice nowadays.

tshirtman (approved):
Didn't test the updated instructions, but indeed it's important to switch to https as much as possible.

obfusk:
I tested the HTTPS URLs to see if they work, though I did not check whether the result is identical.

This is why I added a `www.` prefix to the sphynx URLS, since the URL without the prefix doesn't work over HTTPS.

AndreMiras (approved):
LGTM

obfusk:
Given the security implications and otherwise harmlessness of these changes may I suggest doing a point release that includes (only) these changes?

PR #1328, opened 2021-05-21
Android: Expose support for native services in buildozer

misl6:
This PR exposes support for native services in buildozer.
(see https://github.com/kivy/python-for-android/pull/2244)


AndreMiras (commented):
Looking good thank you.
Do you mind updating the tests?
https://github.com/kivy/buildozer/blob/master/tests/targets/test_android.py

PR #1369, opened 2021-10-12
add the option p4a.extra_args

antocuni:
This option allows to pass arbitrary arguments to p4a.
It is useful especially when using experimental versions and/or forks introducing new options for which buildozer doesn't have a "first class option" yet.

In the following example I used it to use the new experimental `--fileprovider-paths`:
https://github.com/antocuni/plyer_camera_example
https://github.com/antocuni/plyer_camera_example/blob/e820f2bd70795e7c829329135967f751c5ddc706/src/buildozer.spec#L47-L52


antocuni:
> LGTM. Thank you. Can you please fix the conflicts?

done, I rebased on top of the current master

PR #1410, opened 2022-03-13
Update CHANGELOG and prepare release 1.3.0

misl6:
Release time 🥳 

Since a new `python-for-android` release that supports is out, now `buildozer` should support AAB artifacts.

For the reviewer: If the GitHub Actions selects `macos-11` (which is slowly migrating as the default `macos-latest`) macOS tests will likely fail. This is a known issue and will be fixed on `python-for-android` side (along with some tests targeting macOS).


AndreMiras (approved):
LGTM

PR #1423, opened 2022-04-08
Update installation.rst

tct123:
changed "openjdk-13-jdk" to "openjdk-8-jdk", because "openjdk-13-jdk" isn't availible in Ubuntu 21.10.


misl6:
Now `python-for-android` (due to an external dependency)  requires `openjdk-11-jdk` or higher (like `openjdk-17-jdk`)

The documentation has already been updated via #1500

PR #1427, opened 2022-04-11
Changes for NDK23

HyTurtle:
The arch is no longer used in the download url for NDK23+

### DIFF
--- a/buildozer/targets/android.py
+++ b/buildozer/targets/android.py
@@ -474,12 +474,13 @@ class TargetAndroid(Target):
                 ext = 'tar.bz2'
             else:
                 ext = 'zip'
-            archive = 'android-ndk-r{0}-' + _platform + '-{1}.' + ext
+            archive = 'android-ndk-r{0}-' + _platform + '{1}.' + ext
             is_64 = (os.uname()[4] == 'x86_64')
         else:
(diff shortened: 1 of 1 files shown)


misl6:
Nice, thank you!
Considering that the buildozer CI runs on `develop` version of `python-for-android`, can you also please add a patch here so we can see the CI running?
https://github.com/kivy/buildozer/blob/12796343f009a8aad562f31cac2512c8b06557eb/.github/workflows/android.yml#L26

We'll also likely need to open an issue regarding the default NDK version in [default.spec](https://github.com/kivy/buildozer/blob/master/buildozer/default.spec) cause it will need to be changed when a new `python-for-android` version is going to be released.

HyTurtle:
Did you mean modify the default.spec so the sed replacement would no longer be needed in the workflow?

And yea the default ndk is as per python-for-android, if changing that comment in the spec I would also change the target api - is there a consensus there? Was thinking 30; unless p4a already defaults to another

HyTurtle:
That's what was confusing me yea, on the target api - its the 27 used in p4a tests I suppose that's holding things up. Probably wait for that before changing here.

misl6:
Looks that all the relevant tests are passing (macOS Android / Integration is expected to fail).

Issue #1433, opened 2022-04-27
Buildozer silently ignores local recipes – but we're not gonna take it anymore!


leycec:
## tl;dr

Relative dirnames in `p4a.local_recipes` and `--local_recipes` have been broken for a year or two. Until this issue is resolved, please replace the `./` prefixing these dirnames with `../../../../` instead: e.g.,

```
# Instead of this...
p4a.local_recipes = ./p4a-recipes/

# ...just do this!
p4a.local_recipes = ../../../../p4a-recipes/
```

Rejoice as local recipes suddenly work. If you too are now facepalming, this emoji's for you. :facepalm: 

## Versions or It Didn't Happen

* Python: *Any*
* OS: **Gentoo Linux,** of course!
* Buildozer: `1.3.0`

## Exegesis or It Didn't Happen

Buildozer silently ignores relative dirnames for both the `p4a.local_recipes` setting in `buildozer.spec` files *and* the `--local_recipes` option accepted by most `p4a` subcommands. Since the only sane means of specifying a local recipe directory is with a relative dirname, local recipes have effectively been broken for at least a year or two... *which is bad.*

Clearly, local recipes used to work. As this issue will show, they no longer do. The breakage *probably* happened when Buildozer was refactored to change the current working directory (CWD) to the deeply nested `.buildozer/android/platform/python-for-android/` subdirectory before running `p4a`. Previously, Buildozer must not have done that; it must have preserved the CWD as the current top-level project directory containing the root `buildozer.spec` file. Now, Buildozer changes the CWD to a deeply nested subdirectory breaking relative dirnames in `buildozer.spec` files.

Let's dissect this. :scissors: 

### buildozer.spec

Let's use this spec file to build the `Blap` app, Buildozer!

```
[app]
p4a.local_recipes = ./p4a-recipes/
title = Blap
package.name = blap
source.dir = .
version = 0.0.1
requirements = python3>=3.9.0, kivy>=2.1.0, bleak>=0.14.0
android.accept_sdk_license = True
android.archs = arm64-v8a, armeabi-v7a

[buildozer]
log_level = 2
```

Let's define a trivial recipe for the `bleak` requirement that mostly does nothing except exist!

```bash
$ mkdir -p p4a-recipes/bleak
$ cat << EOF > p4a-recipes/bleak/__init__.py
from pythonforandroid.recipe import PythonRecipe

class BleakRecipe(PythonRecipe):
    version = '0.14.0'
[...]

PR #1435, opened 2022-04-30
Avoid the usage of `shell=True` in Popen, which could lead to potential security risks.

misl6:
See full story here:
https://github.com/kivy/buildozer/issues/1433#issuecomment-1112906681
https://github.com/kivy/buildozer/issues/1433#issuecomment-1113074469
https://github.com/kivy/buildozer/issues/1433#issuecomment-1113958798

### DIFF
--- a/buildozer/targets/android.py
+++ b/buildozer/targets/android.py
@@ -32,6 +32,7 @@ from os.path import exists, join, realpath, expanduser, basename, relpath
 from platform import architecture
 from shutil import copyfile, rmtree, which
 import shlex
+import pexpect
 from glob import glob
 from time import sleep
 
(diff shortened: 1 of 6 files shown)


buildozer/targets/android.py
```diff
@@ -569,7 +564,7 @@ def _android_update_sdk(self, *sdkmanager_commands):
             sdkmanager_path = self.sdkmanager_path
             sdk_root = f"--sdk_root={android_sdk_dir}"
             command = f"{yes_command} | {sdkmanager_path} {sdk_root} --licenses"
-            self.buildozer.cmd(command, cwd=self.android_sdk_dir)
+            self.buildozer.cmd(command, cwd=self.android_sdk_dir, shell=True)
```
tshirtman, line 567: i believe this one could be replaced using pipes to capture stdout and inject it in stdin of sdkmanager, but indeed our abstraction probably doesn't make that convenient.

tshirtman (approved):

buildozer/targets/android.py
misl6: Yeah, that was a quick fix in order to see the CI running. I expect to remove it before the merge.

We have another `shell=True` on iOS, that needs some attention, but doesn't make use of anything that is injectable by the user via `buildozer.spec`, so is low-prio.

misl6: Changed this one, but kept the one on iOS.

misl6:
🏁  I think is ready for a review.
✅ I did some additional tests locally and everything looks fine (at least on Android).

misl6:
@tshirtman since you started a review on it, do you see anything that could lock us to merge it?

tshirtman (approved):
Looks good to me, didn't run it, (but tests & CI should catch any issue), and can't see anything wrong with the changes.

PR #1444, opened 2022-06-01
fix aar build

mzakharo:
Fix building aar archives.  Companion CL to this is https://github.com/kivy/python-for-android/pull/2612

### DIFF
--- a/buildozer/target.py
+++ b/buildozer/target.py
@@ -102,7 +102,7 @@ class Target:
     def cmd_debug(self, *args):
         self.buildozer.prepare_for_build()
         self.build_mode = 'debug'
-        self.artifact_format = 'apk'
+        self.artifact_format = self.buildozer.config.getdefault('app', 'android.debug_artifact', 'apk')
         self.buildozer.build()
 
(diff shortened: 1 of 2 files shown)


buildozer/target.py
```diff
@@ -102,7 +102,7 @@ def cmd_update(self, *args):
     def cmd_debug(self, *args):
         self.buildozer.prepare_for_build()
         self.build_mode = 'debug'
-        self.artifact_format = 'apk'
+        self.artifact_format = self.buildozer.config.getdefault('app', 'android.debug_artifact', 'apk')
```
misl6, line 105: Can you also add `android.debug_artifact` to `default.spec` ? (Near to `android.release_artifact` would be nice)

misl6 (changes requested):
Hi @mzakharo !

Left a comment, which should be addressed before merging. (As it's also a nice way to document the feature)

buildozer/target.py
mzakharo: Done

misl6 (approved):
LGTM. Thank you!

PR #1450, opened 2022-06-27
use p4a --add-source instead of manual copy of java files

tito:
Currently, android.add_src does not work anymore.
Using --add-source from p4a make it work again.

### DIFF
--- a/buildozer/targets/android.py
+++ b/buildozer/targets/android.py
@@ -918,6 +918,12 @@ class TargetAndroid(Target):
             cmd.append('--blacklist')
             cmd.append(realpath(expanduser(blacklist_src)))
 
+        # support for java directory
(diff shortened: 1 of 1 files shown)


tito:
I don't think it was about symlink, but buildozer was copying files into a directory that is actually cleaned just before p4a copy its own files. So my files never appeared into the final apk/aab.
I tried to look when it started not working, but didn't found.

At least this approach is using the standard way of p4a, so no problem.

misl6 (approved):
LGTM! Tested and works fine.

Yeah, that's the standard way of `python-for-android`, and there's absolutely no reason to keep a duplicated logic in `buildozer`.

The "old" approach was still working for me, so I was trying to investigate why I didn't encountered the same issue 😄

PR #1462, opened 2022-07-20
Updates default buildozer.spec NDK from 19b to 23b

misl6:
Similar to #1041 

Note that this is only the template spec file and `DEFAULT_ANDROID_NDK_VERSION` fallback is left untouched.


misl6:
PS: Android / Integration (macOS-latest) is expected to fail.

The issue is only present in the CI and has been fixed in https://github.com/kivy/buildozer/pull/1435 (but not merged yet, as it needs some more testing)

PR #1476, opened 2022-07-26
fix target osx

Furtif:
this close https://github.com/kivy/buildozer/pull/1472
 * Unify


onuratakan:
We must give a change for this pr because the osx and ios is some problematic, my issue is show this #1477

Furtif:
the pr solves your problem?

onuratakan:
> the pr solves your problem? @onuratakan

I cant approve now but i will try in next days.

PR #1495, opened 2022-08-16
Fixes some E275 - assert is a keyword.

misl6:
- Assert is a keyword and a space is needed + the parentheses are not needed.
- A recent change in pycodestyle (See: https://github.com/PyCQA/pycodestyle/pull/1063) introduced this check.
- CI runs started to fail this evening.


NomadDemon:
looks ok

Issue #1487, opened 2022-08-11
android.presplash_color bug


pynting:
Not sure if this is a universal problem, but building my app failed when i uncommented the "android.presplash_color" setting in the buildozer.spec file. 

Relevant log:
```bash
[DEBUG]:        > Task :mergeDebugResources FAILED
[DEBUG]:        /.buildozer/android/platform/build-arm64-v8a_armeabi-v7a/dists/kr/src/main/res/values/strings.xml:3:4: Failed to flatten XML for resource 'presplash_color' with error: Invalid unicode escape sequence in string
[DEBUG]:        "{str}"
[DEBUG]:        /.buildozer/android/platform/build-arm64-v8a_armeabi-v7a/dists/kr/src/main/res/values/strings.xml:3:4: string/presplash_color does not contain a valid string resource.
```

And indeed, in the referenced file:

```xml
<?xml version="1.0" encoding="utf-8"?>
<resources>
    <string name="app_name">Kr</string>
    <string name="private_version">c8a851e76193bfc03b01e371acf67e0f938ddcba</string>
    <string name="presplash_color">'aqua'</string>
    <string name="urlScheme">kivy</string>
</resources>
```
Problem lies in the buildozer file android.py:
```python
        # support for presplash background color
        presplash_color = self.buildozer.config.getdefault('app', 'android.presplash_color', None)
        if presplash_color:
            cmd.append('--presplash-color')
            cmd.append("'{}'".format(presplash_color))
```
Changed 
` cmd.append("'{}'".format(presplash_color))`
into 
` cmd.append("{}".format(presplash_color))`
solved the issue for me

### Versions

* Python: 3.10
* OS: Fedora
* Buildozer: 1.4.1.dev0

Olamidipupo-favour:
I had similar issues.

apacelus:
I made a pr to fix this

pynting:
Perfect! Do I close this issue, or should the administrator(s) do it?

apacelus:
> Perfect! Do I close this issue, or should the administrator(s) do it?

Let them do it, or close it after they accept my pr

apacelus:
@pynting my pr was merged, you can close this issue now

misl6:
Closed via #1497

PR #1497, opened 2022-08-24
Fix presplash color

apacelus:
Buildozer wont build if a presplash color is set, this should fix it as described in #1487

### DIFF
--- a/buildozer/targets/android.py
+++ b/buildozer/targets/android.py
@@ -881,7 +881,7 @@ class TargetAndroid(Target):
         presplash_color = self.buildozer.config.getdefault('app', 'android.presplash_color', None)
         if presplash_color:
             cmd.append('--presplash-color')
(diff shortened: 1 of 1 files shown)


misl6 (commented):
I guess we can even switch to `cmd.append(presplash_color)` ?

apacelus:
Maybe, Im not sure though. I just made a pr for the issue

tshirtman:
Using format ensure it’s a string, even if it should always be, there is a possibility (i didn’t check) that it’s detected as a number if it’s an hexadecimal chain. So fine to keep it like this imho.

misl6:
> Using format ensure it’s a string, even if it should always be, there is a possibility (i didn’t check) that it’s detected as a number if it’s an hexadecimal chain. So fine to keep it like this imho.

Absolutely a good point! 👍

PR #1500, opened 2022-08-30
Update installation.rst

RobertFlatt:
Update jdk and Ubuntu versions.
Update WSL usage instructions.

### DIFF
--- a/docs/source/installation.rst
+++ b/docs/source/installation.rst
@@ -41,7 +41,7 @@ After installing WSL and Ubuntu on your Windows machine, open Ubuntu, run the co
 
 Copy your Kivy project directory from the Windows partition to the WSL partition, and follow the Quickstart Instructions. **Do not** change to the project directory on the Windows partition and build there, this may give unexpected and obscure fails. 
 
-For debugging, WSL does not have direct access to USB. Copy the .apk file to the Windows partition and run  ADB (Android Debug Bridge) from a Windows prompt. ADB is part of Andriod Studio, if you do not have this installed you can install just the platform tools which also contain ADB. 
+For debugging, WSL does not have direct access to USB. Copy the .apk file to the Windows partition and run ADB (Android Debug Bridge) from a Windows prompt. ADB is part of Android Studio, if you do not have this installed you can install just the platform tools which also contain ADB. 
 
 - Go to https://developer.android.com/studio/releases/platform-tools and click on "Download SDK Platform-Tools for Windows".
 
--- a/docs/source/installation.rst
+++ b/docs/source/installation.rst
@@ -37,7 +37,7 @@ To use buildozer in Windows you need first to enable Windows Subsystem for Linux
 
 These instructions were tested with WSL 1 and Ubuntu 18.04 LTS, and WSL2 with Ubuntu 20.04 and 22.04. 
 
-After installing WSL and Ubuntu in your Windows machine, open Ubuntu and run the commands listed on the previous section and restart your WSL terminal to enable the path change.
+After installing WSL and Ubuntu on your Windows machine, open Ubuntu, run the commands listed in the previous section, and restart your WSL terminal to enable the path change.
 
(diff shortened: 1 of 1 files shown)


docs/source/installation.rst
```diff
@@ -14,49 +14,34 @@ First, install the buildozer project with::
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-To use buildozer in Windows 10 you need first to enable Windows Subsystem for Linux (WSL) and install a Linux distribution: https://docs.microsoft.com/en-us/windows/wsl/install-win10.
+To use buildozer in Windows you need first to enable Windows Subsystem for Linux (WSL) and install a Linux distribution: https://docs.microsoft.com/en-us/windows/wsl/install.
 
-These instructions were tested with WSL 1 and Ubuntu 18.04 LTS. 
+These instructions were tested with WSL 1 and Ubuntu 18.04 LTS, and WSL2 with Ubuntu 20.04 and 22.04. 
 
-After installing WSL and Ubuntu in your Windows 10 machine, open Ubuntu and do this:
+After installing WSL and Ubuntu in your Windows machine, open Ubuntu and run the commands listed on the previous section and restart your WSL terminal to enable the path change.
```
misl6, line 40:
```suggestion
After installing WSL and Ubuntu on your Windows machine, open Ubuntu, run the commands listed in the previous section, and restart your WSL terminal to enable the path change.
```

docs/source/installation.rst
```diff
@@ -14,49 +14,34 @@ First, install the buildozer project with::
...
-    source venv/bin/activate
-    python -m pip install --upgrade pip
-    python -m pip install --upgrade wheel
-    python -m pip install --upgrade cython 
-    python -m pip install --upgrade virtualenv 
-    python -m pip install --upgrade buildozer
-    # Restart your WSL terminal to enable the path change
-
-Windows Subsystem for Linux does not have direct access to USB. Due to this, you need to install the Windows version of ADB (Android Debug Bridge):
+For debugging, WSL does not have direct access to USB. Copy the .apk file to the Windows partition and run  ADB (Android Debug Bridge) from a Windows prompt. ADB is part of Andriod Studio, if you do not have this installed you can install just the platform tools which also contain ADB.
```
misl6, line 44:
```suggestion
For debugging, WSL does not have direct access to USB. Copy the .apk file to the Windows partition and run ADB (Android Debug Bridge) from a Windows prompt. ADB is part of Android Studio, if you do not have this installed you can install just the platform tools which also contain ADB. 
```

misl6 (commented):
I Left 2 minor suggestions, but absolutely LGTM. Thank you! ❤️ 

I'm wondering if it's better to also include a section that advises using a full-featured VM, instead of WSL*, but that's another topic. 🧐

RobertFlatt:
Thank you for the improvements.

One question, is Cython still a user install requirement? I ask because somewhere I noticed Buildozer using Cython 3 - which means it downloads its own version. Can Cython be removed here?

Yes, any VE documentation should be platform independent. (IMO for somebody who installed WSL just to run Buildozer, a VE is complexity without utility - which is why I removed it).

misl6:
> One question, is Cython still a user install requirement? I ask because somewhere I noticed Buildozer using Cython 3 - which means it downloads its own version. Can Cython be removed here?

If Cython doesn't get installed by the user `buildozer` will fail here ( something similar happens on the iOS side ):
https://github.com/kivy/buildozer/blob/1a7191e0e456c6e6efc721a07b31c3becdc32126/buildozer/targets/android.py#L297
And nope, `python-for-android` doesn't install (or give instructions to `buildozer` via install reqs ) Cython.

`python-for-android` automatically installs Cython only for regular packages installations here:

https://github.com/kivy/python-for-android/blob/28151d1864df9564ee796910f38a995b54d73631/pythonforandroid/build.py#L715

(Previously [~3/4 years ago ] we were using the system-wide cython install, so that's why)

Yeah, we may need to do some renovations 🏡 , but now is still a user install requirement 😄.

> Yes, any VE documentation should be platform independent. (IMO for somebody who installed WSL just to run Buildozer, a VE is complexity without utility - which is why I removed it).

Agree!

misl6:
Thanks again!

misl6 (approved):

PR #1512, opened 2022-10-10
Update Defaults

RobertFlatt:
Update defaults to

android.api = 31
android.enable_androidx = True

### DIFF
--- a/buildozer/default.spec
+++ b/buildozer/default.spec
@@ -99,7 +99,7 @@ fullscreen = 0
 #android.features = android.hardware.usb.host
 
 # (int) Target Android API, should be as high as possible.
-#android.api = 27
+#android.api = 31
 
 # (int) Minimum API your APK / AAB will support.
 #android.minapi = 21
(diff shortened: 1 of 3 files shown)


RobertFlatt:
Are these fails an artifact of the test environment?

misl6:
> Are these fails an artifact of the test environment?

iOS: 100%  an artifact
Android: `tests` need to be updated accordingly to your changes.

misl6 (changes requested):
Can you please update the tests and ping me back when you're ready?

RobertFlatt:
@misl6  ping

buildozer/default.spec
```diff
@@ -192,7 +192,7 @@ fullscreen = 0
 # (bool) Enable AndroidX support. Enable when 'android.gradle_dependencies'
 # contains an 'androidx' package, or any package from Kotlin source.
 # android.enable_androidx requires android.api >= 28
-#android.enable_androidx = False
+android.enable_androidx = True
```
misl6, line 195: How about leaving the option commented (but with the updated default):
```suggestion
#android.enable_androidx = True
```

and having the default changed here?

https://github.com/kivy/buildozer/blob/b8575fd5b1e150bf93be3b7e77f229a28a7e5e0d/buildozer/targets/android.py#L1265-L1267

That will allow us to enable it by default, even for current users that may have an outdated `buildozer.spec`

misl6 (changes requested):
Left a suggestion regarding `androidx`, let me know what you think about it 😃

RobertFlatt:
@misl6 Good idea, implemented.

PR #1513, opened 2022-10-10
Add resources

RobertFlatt:
Complement to p4a PR

### DIFF
--- a/buildozer/default.spec
+++ b/buildozer/default.spec
@@ -186,6 +186,12 @@ fullscreen = 0
 # 2) android.add_assets = source_asset_path:destination_asset_relative_path
 #android.add_assets =
 
+# (list) Put these files in the apk res directory.
+# Either form may be used, and resources need not be in 'source.include_exts'.
+# 1) android.add_resources = source_resource_relative_path
+# 2) android.add_resources = source_resource_path:destination_resource_relative_path
+#android.add_resources =
+
 # (list) Gradle dependencies to add
 #android.gradle_dependencies =
 
--- a/buildozer/targets/android.py
+++ b/buildozer/targets/android.py
@@ -931,6 +931,17 @@ class TargetAndroid(Target):
                 asset_dest = asset
             cmd.append(realpath(expanduser(asset_src)) + ':' + asset_dest)
 
+        # support for res folder
+        resources = self.buildozer.config.getlist('app', 'android.add_resources', [])
+        for resource in resources:
+            cmd.append('--add-resource')
+            if ':' in resource:
+                resource_src, resource_dest = resource.split(":")
+            else:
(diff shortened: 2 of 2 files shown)


misl6:
For the logs:
Related to https://github.com/kivy/python-for-android/pull/2684, so we should keep this PR on hold.

buildozer/targets/android.py
```diff
@@ -931,6 +931,17 @@ def execute_build_package(self, build_cmd):
...
+        # support for res folder
+        resources = self.buildozer.config.getlist('app', 'android.add_resources', [])
+        for resource in resources:
+            cmd.append('--add-resource')
+            if ':' in resource:
+                resource_src, resource_dest = resource.split(":")
+            else:
+                resource_src = resource
+                resource_dest = ""
+            cmd.append(realpath(expanduser(resource_src)) + ':' + resource_dest)
```
misl6, line 943: Ouch, duplicating such logic on `buildozer` may lead to something unexpected (E.g: I needed to recall how the no `:` syntax worked on `python-for-android` side.)

This is bad, but I'm NOT blaming you, since we're expanding paths on `buildozer` side everywhere. ( Bad, and should be moved on `python-for-android` ? )

buildozer/default.spec
```diff
@@ -186,6 +186,18 @@ fullscreen = 0
...
 #android.add_assets =
 
+# (list) Put these files or directories in the apk res directory.
+# The option may be used in three ways, the value may contain one or zero ':'
+# Some examples:
+# 1) A file to add to resources, legal resource names contain ['a-z','0-9','_']
+# android.add_resources = my_icons/all-inclusive.png:drawable/all_inclusive.png
+# 2) A directory, here  'legal_icons' must contain resources of one kind
+# android.add_resources = legal_icons:drawable
+# 3) A directory,here 'legal_resources' must contain one or more directories,
```
misl6, line 196:
```suggestion
# 3) A directory, here 'legal_resources' must contain one or more directories, 
```

misl6 (changes requested):
Left a super-minor suggestion (typo) and a thought about something that you should not address ATM, but that we should care about if we want to avoid to break the user experience at some point in the future.

buildozer/targets/android.py
RobertFlatt: Yes, absolutely, but as you point out this is consistent with for example https://github.com/kivy/buildozer/blob/master/buildozer/targets/android.py#L931L938

I choose to copy what is shown to work, it is the low risk implementation. And yes the whole interface needs a rewrite, and an automatic appclean, and the .spec reorganized into common and uncommon, and....

misl6 (approved):
LGTM. Thank you!

PR #1516, opened 2022-10-14
CeDeROM add FreeBSD support 20221014.

cederom:
1. Initial FreeBSD support using Linux binaries.
* FreeBSD can emulate Linux ELF binaries [1].
* Use the same process for FreeBSD as for Linux.
* Add FreeBSD recognition in checks.

[1] https://docs.freebsd.org/en/articles/linux-emulation/

2. Fix 64-bit detection based on os.uname.
* Not every OS uses `x86_64` in `uname` (i.e. FreeBSD uses `amd64`).
* Every 64-bit OS will have 64 in `uname`.
* Alternatively `platform.architecture()[0]` will return `64bit`.

### DIFF
--- a/buildozer/targets/android.py
+++ b/buildozer/targets/android.py
@@ -474,7 +474,7 @@ class TargetAndroid(Target):
             else:
                 ext = 'zip'
             archive = 'android-ndk-r{0}-' + _platform + '{1}.' + ext
-            is_64 = (os.uname()[4] == 'x86_64')
+            is_64 = ('64' in os.uname()[4])
         else:
             raise SystemError('Unsupported platform: {}'.format(platform))
 
--- a/buildozer/targets/android.py
+++ b/buildozer/targets/android.py
@@ -265,7 +265,7 @@ class TargetAndroid(Target):
                                 'adb.exe')
             self.javac_cmd = self._locate_java('javac.exe')
             self.keytool_cmd = self._locate_java('keytool.exe')
-        # darwin, linux
+        # darwin, linux, freebsd
         else:
             self.adb_executable = join(self.android_sdk_dir, 'platform-tools', 'adb')
             self.javac_cmd = self._locate_java('javac')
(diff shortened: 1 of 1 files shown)


buildozer/targets/android.py
```diff
@@ -279,6 +279,10 @@ def check_requirements(self):
                     'zlib headers must be installed, '
                     'run: sudo apt-get install zlib1g-dev')
 
+            # Set environment to use Linux binaries on FreeBSD at later stages.
```
misl6, line 282:
```suggestion
            # Overrides the OS which `sdkmanager` should download the packages for.
            # See: https://developer.android.com/studio/command-line/variables#envar
```

misl6 (commented):
LGTM.
Left a minor suggestion that may help contributors in future.

buildozer/targets/android.py
cederom: THanks @misl6 I have updated the description with your hints and why this is used :-)

cederom: Sorry, i amended that to 64-bit detection commit, fix incoming :-)

cederom: rebased :-)

cederom:
1. Provided patches enables download and run of Linux binaries on FreeBSD.
2. Google NDK also has some Linux only hardcodes, but fixing them enables build on FreeBSD. Problem already reported to NDK Dev Team: https://github.com/android/ndk/issues/1780#issuecomment-1281390278.
3. Some Recipes can be built and some not because of Linux only hardcodes, but this is a separate P4A issue, probably will need a dedicated fix to each of the problematic dependency upstream. TODO.

cederom:
Here is my local fork of NDK with several commits that enables build on FreeBSD. Bugs are reported to Google Android NDK team + fixes provided.

https://github.com/cederom/google-android-ndk/tree/cederom-freebsd

cederom:
Not sure if its worth the effort to work with a so called "OpenSource" project (Google/Android/NDK) that rejects (as "not planned") even trivial patches to make it work on other platforms than officially supported :-(

https://github.com/android/ndk/issues/1785

cederom:
Allright two of the three patches were accepted :-)

* https://android-review.googlesource.com/c/platform/ndk/+/2261302
* https://android-review.googlesource.com/c/platform/ndk/+/2261304

Third one is in progress and under discussion on how to best enable FreeBSD operations without having official support from Google NDK team :-)

* https://android-review.googlesource.com/c/platform/ndk/+/2261303

misl6 (approved):
LGTM. Thank you!
Please keep us posted about FreeBSD support on Android NDK. 😄

cederom:
Thank you @misl6 :-) When is next Buildozer release planned? :-)

misl6:
> Thank you @misl6 :-) When is next Buildozer release planned? :-)

Nothing planned, but considering the current status we are likely going to have a new release in the first week of November.

Issue #1535, opened 2022-12-13
android.extra-manifest entry inserts bad code into AndroidManifest


lufebe16:
### Versions

* Python: 3.10
* OS: any
* Buildozer: newest pip version

### Description

using the entry for:

 android.extra_manifest_xml = ./src/manifest/extra_permission.xml

to add a new permission. The file contains:
`
<uses-permission android:name="android.permission.READ_EXTERNAL_STORAGE"/>
`
This results in a build failure on the last step. Gradle reclaims an error on parsing AndroidManifest.xml.

Investigation of the problem showed, that buildozer obviously removed all double quotes from the xml entry that has been added.

### Correction

To correct this failure one needs to change the following code in the source (android.py). Hier is a patch that applies to the current master.

[extra.zip](https://github.com/kivy/buildozer/files/10218375/extra.zip)

RobertFlatt:
As a workaround use `android.permissions=READ_EXTERNAL_STORAGE` in buildozer.spec

misl6:
Closed via #1555

PR #1537, opened 2022-12-23
Corrections (fixes) for

lufebe16:
- extra-manifest-xml
- extra-nanifest-application-attributes

This	also refers to issue #1535

My build has now been tested against the newest p4a and buldozer versions (cloned buildozer/master and python-for-android/develop). It fails for extra-manifest-xml and also for extra-manifest-application-attributes.

Obviously the new version of p4a behaves differently regarding processing of double quotes in Android Manifest	templates. Needs the buildozer code 
to reflect that.


misl6:
Duplicate of #1555

PR #1549, opened 2023-01-22
Removes broken and unneeded checks, now we can support additional permission syntaxes.

misl6:
Related python-for-android PR: https://github.com/kivy/python-for-android/pull/2725

`android.permissions` now accepts all the syntaxes supported by `python-for-android`.

Removed broken and un-needed checks related to android permissions:

- `_get_available_permissions ` is looking for a file that does not exist anymore at the provided location, so everything is already considered "ok". Basically the check is already skipped.
- Forcing the last component to be uppercase can lead to multiple issues.

In general, IMHO is better to leave the error-handling for Android-specific things to the Android toolchain (or to `python-for-android` if it makes sense), as these things change fast .

### DIFF
--- a/buildozer/default.spec
+++ b/buildozer/default.spec
@@ -93,7 +93,8 @@ fullscreen = 0
 #icon.adaptive_background.filename = %(source.dir)s/data/icon_bg.png
 
 # (list) Permissions
-#android.permissions = INTERNET
+# (See https://python-for-android.readthedocs.io/en/latest/buildoptions/#build-options-1 for all the supported syntaxes and properties)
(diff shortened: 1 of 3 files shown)


AndreMiras (approved):
LGTM
One thing, the commit description says "Allow additional control around Android permissions", but I don't see the related change

misl6:
> LGTM
> One thing, the commit description says "Allow additional control around Android permissions", but I don't see the related change

Changed to "Removes broken and unneeded checks, now we can support additional permission syntaxes."
For the records: The broken and unneeded checks were preventing the chance to use the additional permission syntaxes.

PR #1555, opened 2023-01-30
Corrections (fixes) for extra-manifest-xml

lufebe16:
Corrections for:
    - extra-manifest-xml
    - extra-nanifest-application-attributes

following issues detected:

android.features = android.hardware.sensor.accelerometer
results in:

toolchain.py: error: unrecognized arguments: --feature android.hardware.sensor.accelerometer

as a workaround I tried to supply a file containing the xml like:

android.extra_manifest_xml = ./xml/feature.xml

then toolchain parameter generated bei buildozer is wrong.
`... , '--extra-manifest-xml="<uses-feature\n      android:name="android.hardware.sensor.accelerometer"\n        android:required="true" />\n"', ...
`
p4a adds all contents after the equal sign to AndroidManifest, i.e including these additional double quotes - this is not correct.

After my change this works correctly. It produces the toolchainparameter (in 2 pieces) as:

`... , '--extra-manifest-xml', '<uses-feature\n        android:name="android.hardware.sensor.accelerometer"\n        android:required="true" />\n', ...
`
The same change is needed for extra_manifest_application_arguments.

This also fixes issue #1535.

### DIFF
--- a/buildozer/targets/android.py
+++ b/buildozer/targets/android.py
@@ -884,14 +884,15 @@ class TargetAndroid(Target):
         extra_manifest_xml = self.buildozer.config.getdefault(
             'app', 'android.extra_manifest_xml', '')
         if extra_manifest_xml:
-            cmd.append('--extra-manifest-xml="{}"'.format(open(extra_manifest_xml, 'rt').read()))
+            cmd.append('--extra-manifest-xml')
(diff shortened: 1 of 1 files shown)


misl6 (approved):
LGTM. Thank you!

PR #1576, opened 2023-03-06
display-cutout

kuzeyron:
This PR enables display-cutout functionality.

### DIFF
--- a/buildozer/default.spec
+++ b/buildozer/default.spec
@@ -368,7 +368,7 @@ ios.kivy_ios_branch = master
 #ios.ios_deploy_dir = ../ios_deploy
 # Or specify URL and branch
 ios.ios_deploy_url = https://github.com/phonegap/ios-deploy
-ios.ios_deploy_branch = 1.10.0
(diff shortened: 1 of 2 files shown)


kuzeyron:
Final changes.

mmdmoa:
bro that's so cool, I am craving this feature for a long time now!

mmdmoa:
by the way, what do you mean by enabling display-cutout functionality?
do you mean edge to edge display rendering?

kuzeyron:
> by the way, what do you mean by enabling display-cutout functionality? do you mean edge to edge display rendering?

I think the answer is yes. Without this function Kivy doesn't render beyond the status-bar even when it's set to fullscreen so this gives you the option to render more on the display.

mmdmoa:
> > by the way, what do you mean by enabling display-cutout functionality? do you mean edge to edge display rendering?
> 
> I think the answer is yes. Without this function Kivy doesn't render beyond the status-bar even when it's set to fullscreen so this gives you the option to render more on the display.

That's great, I'm so waiting for this feature to be out! thanks!

PR #1600, opened 2023-05-08
Update ios_deploy branch to 1.12.2 Needed for latest Xcode.

akshayaurora:
Latest Xcode removes support for older simulators and complains while trying to link

```
ld: file not found: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/lib/arc/libarclite_macosx.a
```

Using latest branch from ios_deploy seems to fix the issue.


misl6:
FYI @akshayaurora (If you haven't already started an investigation), the tests are failing as `1.10.0` is expected during the clone.

PR #1626, opened 2023-07-06
Update default.spec

RobertFlatt:
In buildozer.spec

> \# (str) python-for-android git clone directory (if empty, it will be automatically cloned from github)
> \# p4a.source_dir = ../python-for-android

p4a is not automatically cloned.  Log message is 

> \# Install platform
> \# Failed to read python-for-android setup.py at ../python-for-android/setup.py

Change doc string to 

> \# (str) python-for-android git clone directory


akshayaurora (approved):

akshayaurora:
👍

PR #1628, opened 2023-07-06
Refactor checkbin to use shutil.which

Julian-O:
checkbin used custom code to find where a binary is on the path. 
shutil.which is a built-in to do the same thing since Python 3.3. 
Simpler and works across platforms.


Julian-O:
Failed tests are for Python 2, that is no longer supported.

misl6:
> Failed tests are for Python 2, that is no longer supported.

Yeah, these checks are only here to make sure we do not remove the "Unsupported Python version" statement.
But, Python 2 is not even supported anymore by Github Actions `action/setup-python`.

I guess we can remove this check, as there's no need to over-complicate things here.

misl6 (approved):
LGTM. Thank you!

PR #1632, opened 2023-07-08
Factor out logging code from Buildozer class

Julian-O:
First step toward #1629.

Code related to logging moved into its own class/file.
References (including in targets and tests) updated.
Unit tests separated out, and enhanced to test more conditions.
Trivial indenting/grammar issues in osx.py fixed.
No-op exception handling code removed.

### DIFF
--- a/buildozer/targets/osx.py
+++ b/buildozer/targets/osx.py
@@ -17,15 +17,15 @@ class TargetOSX(Target):
     targetname = "osx"
 
     def ensure_sdk(self):
-        self.buildozer.info('Check if kivy-sdk-packager exists')
+        self.logger.info('Check if kivy-sdk-packager exists')
         if exists(
                 join(self.buildozer.platform_dir, 'kivy-sdk-packager-master')):
-            self.buildozer.info(
+            self.logger.info(
                     'kivy-sdk-packager found at '
                     '{}'.format(self.buildozer.platform_dir))
             return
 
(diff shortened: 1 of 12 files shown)


buildozer/targets/osx.py
```diff
@@ -20,14 +22,14 @@ def ensure_sdk(self):
                 join(self.buildozer.platform_dir, 'kivy-sdk-packager-master')):
             self.buildozer.info(
```
Julian-O, line 23: Missed this one.

misl6 (changes requested):
Hi @Julian-O !

In `buildozer/scripts/remote.py` references have not been updated.
Can you please take care of it?

misl6 (changes requested):
Seems we still have two old references:
https://github.com/kivy/buildozer/blob/0851a101b5e301803a42ed5f5562e6910bff116b/buildozer/scripts/client.py#L19
https://github.com/kivy/buildozer/blob/0851a101b5e301803a42ed5f5562e6910bff116b/buildozer/scripts/remote.py#L274

Unfortunately I missed them from the previous review.
(That part will likely need a huge cleanup, but makes sense to keep it updated ATM)

Julian-O:
These sneaky ones created a Buildozer() object just to call logging! I missed them in my text searches. That left me red-faced.  I should have caught these, sorry.

misl6 (approved):
LGTM. Thank you!

(The CI pipeline is failing, but is a known issue with Cython 3.0.0)

PR #1635, opened 2023-07-16
Remove `python2` test as `python2` is not available anymore via `actions/setup-python`

misl6:
This test was only meant to make sure that we show the "not compatible with python2" error.
More testing is generally great, but unfortunately `python2` is not available anymore via `setup-python` and the pipeline is now failing.

If we want to keep the test up and running, we need to do that in another way, but it will increase the complexity of our CI.


akshayaurora (approved):

akshayaurora:
👯

Issue #1636, opened 2023-07-20
ConfigParser code should be separated


Julian-O:
### Versions

* Buildozer: 1.5.1

### Description

[I am preparing a PR for this, but it is blocked by other higher-priority ones, so I am documenting the plan as a placeholder.]

The Buildozer class has a `config` object that is used by itself and the targets. It is a variant of the Python-standard `ConfigParser`, that supports case-sensitivity, lists and env variable overrides.

It has a number of minor issues:

* The responsibility of parsing the config doesn't need to be in the class Buildozer. (See #1629)
* New methods are added (and overridden) via monkey-patching on the instance, where a simple subclass would suffice.
* There are no unit tests (probably because above decisions make them difficult).
* Inconsistent method naming.
* Config specs are made case-sensitive, which might be necessary but no justification is given. (Not that Env Variable overrides are case-insensitive.)
* Reimplements existing functionality (ConfigParser already supports defaults as "fallbacks")
* Inconsistent use of defaults (sometimes they are parsed, sometimes they are returned as is).
* Environment variable overrides are applied at construction time and then reapplied on each access.
* List-handling is re-implemented several times (with different parameters defined).
* The term "raw" is used in a method differently to how ConfigParser uses it (untrimmed values versus uninterpolated values).

# Proposal

In a separate file (`specparser.py`), define SpecParser, a subclass of ConfigParser.

Implement a single `getlist` method with parameters to support defaults (returned as is), fetching "values", stripping, section separators and split characters.

Add convenience functions that map existing calls to getlist/get/getboolean with appropriate parameters. Deprecate them, because they are unnecessary.

Apply the env var overrides automatically, after the text is read and parsed, before it is accessed. (Note: This means changes to the environment variables while Buildozer is running won't take effect on the config - but I haven't found any examples of that. One could just set the config option directly in these cases.)

Remove the term "raw".

Retain case-sensitivity until it can be shown to be unnecessary.

Write unit-tests.

Update Buildozer class to instantiate the new class rather than ConfigParser. Remove all the monkey patching and the now unnecessary call to set env var overrides.

Update android.py to use getlist with appropriate params rather than the "raw" function.

RobertFlatt:
FYI, in addition here are a couple of buildozer.spec parser issues that result in non-deterministic behavior, and trip up users:

 - There are Python style trailing comments in the buildozer.spec file, or comment characters inside a list. Comment characters must be the first character on a line in the buildozer.spec file.

[...]

PR #1639, opened 2023-07-21
Separate ConfigSpec parser (Address #1636)

Julian-O:
Separate ConfigParser code from Buildozer.

- Use subclass, rather than monkey-patching.
- Single implementation of list-handling code.
- Convenience functions to minimise impact of existing code (but instantly deprecated).
- Treat defaults consistently
- Apply the env overrides automatically (client doesn't need to trigger).
- Apply the env overrides once (read-time) rather than on every get.
- Avoid re-using term "raw" which has pre-existing definition in ConfigSpec.
   - Update android.py client to match,
- Unit tests.

Add comments to start and end of default.spec to explain the syntax (as discussed in #1636)

### DIFF
--- a/buildozer/__init__.py
+++ b/buildozer/__init__.py
@@ -76,7 +76,6 @@ class Buildozer:
         self.specfilename = filename
         self.state = None
         self.build_id = None
-        self.config_profile = ''
         self.config = SpecParser()
 
         self.logger = Logger()
@@ -900,6 +899,8 @@ class Buildozer:
         self.run_command(cmd)
 
     def run_command(self, args):
+        profile = None
+
         while args:
             if not args[0].startswith('-'):
                 break
@@ -913,13 +914,13 @@ class Buildozer:
                 exit(0)
 
             elif arg in ('-p', '--profile'):
-                self.config_profile = args.pop(0)
+                profile = args.pop(0)
 
             elif arg == '--version':
                 print('Buildozer {0}'.format(__version__))
                 exit(0)
 
-        self._merge_config_profile()
+        self.config.apply_profile(profile)
 
         self.check_root()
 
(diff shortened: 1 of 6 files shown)


misl6:
Hi @Julian-O ,
Sorry to bother you, but can you please rebase on top of the latest `master` so we can have a clean CI build? 😀

Julian-O:
@misl6: Delighted to. I am feeling more comfortable now the main tests are clearly passing. (At the time of writing, the slower ones are still running.)

Julian-O:
Tests are all green! Yay

misl6:
> @misl6: Delighted to. I am feeling more comfortable now the main tests are clearly passing. (At the time of writing, the slower ones are still running.)

Seems that the commit has been added, instead of a rebase, so it shows up in the changes. You can rebase instead?

Julian-O:
@misl6: I think that's fixed. Still learning a new git front-end.

misl6:
What are your thoughts about `_merge_config_profile` ?

Julian-O:
My thoughts? A little deflated at the moment. :-)

The first time I looked at it, I noticed it was not treated monkey patched. I was mislead by the name into thinking it was merging two configs, and I thought it had build-specific logic in there, so I wanted to keep it in Buildozer.

I see now it is more general than that, and belongs in SpecParser...

... but it should be called implicitly. The clients shouldn't need to know when to call it. I think the profile name should be passed to the SpecParser's constructor so it can handle this... 

But that isn't available until after the command line arguments are parsed. This is entirely consistent with my longer-term view that Buildozer should be split into BuildozerCommand (which is responsible for handles all the parsing of the options, and help) and Builder (which is responsible for knowing how to build an app). The Builder should be constructed by BuildozerCommand and give it all the parameters it needs in the constructor - including the Profile name, so it can construct the SpecParser.

But that is detouring me more and more from my primary goal. I want to get to the "move BuildOps into its own class, and make them platform independent". I am holding back on that PR while I set the groundwork.

Let me think on this. I will try to find a compromise - like moving `_merge_config_profile` into `SpecParser`, but requiring the clients to explicitly call it, and leave fixing that to the future.

Julian-O:
How about that?

misl6:
> How about that?

Love that.

misl6 (approved):
LGTM. Thank you!

PR #1648, opened 2023-07-31
Minor Win32 improvements

Julian-O:
There is no win32 support for the Android target. This doesn't change that. It does, however, enable experimental support, via an (unpublished) environment variable, to allow exploration of what is required.

Also, there are some signs of vestigial attempts at supporting Windows. As they haven't been run before, they need a little care:

- Updated references from _winreg to winreg (Built-in module was renamed in Python 3.)
- Corrected download URL to Android NDK for the Windows platform
- Supported sdkmanager installer having a different name (.bat extension) on Windows.

- Also corrected trivial formatting error in Exception text, which applies to all platforms.

### DIFF
--- a/buildozer/targets/android.py
+++ b/buildozer/targets/android.py
@@ -2,8 +2,9 @@
 Android target, based on python-for-android project
 '''
 
+import os
 import sys
-if sys.platform == 'win32':
+if sys.platform == 'win32' and not os.getenv("KIVY_WIN32_ANDROID_EXPERIMENTAL"):
     raise NotImplementedError('Windows platform not yet working for Android')
 
 from platform import uname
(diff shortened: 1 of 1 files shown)


buildozer/targets/android.py
```diff
@@ -245,11 +245,14 @@ def apache_ant_dir(self):
     @property
     def sdkmanager_path(self):
         sdkmanager_path = join(
-            self.android_sdk_dir, 'tools', 'bin', 'sdkmanager')
+            self.android_sdk_dir, 'tools', 'bin',
+            'sdkmanager.bat'
+            if platform in ('win32', 'cygwin')
+            else 'sdkmanager')
```
misl6, line 251: Just for readability, how about defining the executable (`sdkmanager` or `sdkmanager.bat`) outside of the `join()`?

Julian-O: Good idea. Done.

misl6 (approved):
LGTM. Thank you!

PR #1653, opened 2023-08-03
Document service options in default.spec

Julian-O:
Also trivial copy-edits to other comments.

### DIFF
--- a/buildozer/default.spec
+++ b/buildozer/default.spec
@@ -59,6 +59,7 @@ requirements = python3,kivy
 orientation = portrait
 
 # (list) List of services to declare
+# This is currently only relevant to Android services.
 # Each service consists of a name (a valid Java class name, with the first letter capitalized)
 # followed by a colon, followed by the name of the Python script (.py file) that should be
 # launched. This is optionally followed by ":foreground" for foreground services or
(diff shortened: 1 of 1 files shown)


Julian-O:
Slipped in another comment to explain how to find what bootstrap options were valid.

buildozer/default.spec
```diff
@@ -58,7 +58,12 @@ requirements = python3,kivy
 # Valid options are: landscape, portrait, portrait-reverse or landscape-reverse
 orientation = portrait
 
-# (list) List of service to declare
+# (list) List of services to declare
+# Each service consists of a name (a valid Java class name, with the first letter capitalized)
+# followed by a colon, followed by the name of the Python script (.py file) that should be
+# launched. This is optionally followed by ":foreground" for foreground services or
+# ":foreground:sticky" for sticky foreground services. The default is a background service.
+# Bound services are not supported.
```
misl6, line 66: Maybe we want to say that (at least ATM) this is an android-specific feature, even if does not have the `android.*` prefix?

Julian-O: Ah, I hadn't even noticed it didn't have an android prefix. I am on it.

misl6 (approved):
LGTM. Thank you!

PR #1668, opened 2023-08-23
Migrate all usages of mkdir and rmdir to buildops

Julian-O:
This is a refactor as one step to reduce the size/complexity of the Buildozer class. It doesn't change functionality. except to improve logging consistency and a small performance improvement

* mkdir() and rmdir() methods removed from `buildozer/__init__.py` (cleaner versions are already in `buildops.py`)
* All references to buildozer's mkdir and rmdir changed over to use buildops.
* Call to Linux's mkdir via a separate shell replaced with (faster, and platform independent) library call.
* Re-ordered imports, where touched, to match PEP8.

### DIFF
--- a/buildozer/__init__.py
+++ b/buildozer/__init__.py
@@ -8,21 +8,21 @@ Generic Python packager for Android / iOS. Desktop later.
 
 __version__ = '1.5.1.dev0'
 
-import os
-import re
-import sys
-import select
 import codecs
+from copy import copy
+from fnmatch import fnmatch
+import os
+from os import environ, unlink, walk, sep, listdir
+from os.path import join, exists, dirname, realpath, splitext, expanduser
+import re
+from re import search
+import select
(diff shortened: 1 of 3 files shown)


buildozer/__init__.py
```diff
@@ -361,22 +362,22 @@ def check_build_layout(self):
             exit(1)
 
         # create global dir
-        self.mkdir(self.global_buildozer_dir)
-        self.mkdir(self.global_cache_dir)
+        mkdir(self.global_buildozer_dir)
```
misl6, line 365: To help people understand that `mkdir` (and other methods) are from `buildops`, how about importing `buildops` and than using `buildops.mkdir`?

Yeah, is quite verbose, but IMHO, the readability (and the understandability) gets improved a lot.

What do you think about it?

Julian-O:
I don't mind either way. Will try to get an updated version to you tomorrow.

PR #1669, opened 2023-08-25
Migrate to use Buildop's rmtree, file_copy, file_copytree and rename

Julian-O:
This is part of a bigger refactor to reduce the size and complexity of the Buildozer class. It adds no new functionality (apart from making logging more consistent)

* Remove file_rename, file_copy and file_copytree from Buildozer.
* Update all references to those Buildozer methods to refer to buildops equivalents.
   * Note: `file_rename` is now called rename as, in practice, it was also used to rename directories.
* Updated all references to shutil.copyfile and shutil.rmtree to use the Buildops equivalents
  * Note this adds implicit debug logging, making them more consistent.
* Updated all references to check_call() and check_output() that called the `rm` and `cp` shell commands.
  * Note this adds implicit debug logging, making them more consistent, and improves performance.
* Updated mock in test.

Cheekily, also removed some obsolete references to Python 2 (only) standard libraries to make my IDE stop warning me about them while I edited the file.

### DIFF
--- a/buildozer/__init__.py
+++ b/buildozer/__init__.py
@@ -17,7 +17,7 @@ from os.path import join, exists, dirname, realpath, splitext, expanduser
 import re
 from re import search
 import select
-from shutil import copyfile, rmtree, copytree, move, which
+from shutil import which
 from subprocess import Popen, PIPE, TimeoutExpired
 import sys
 from sys import stdout, stderr, exit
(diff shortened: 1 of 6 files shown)


buildozer/__init__.py
```diff
@@ -476,23 +476,6 @@ def file_matches(self, patterns):
     def file_exists(self, *args):
         return exists(join(*args))
 
-    def file_rename(self, source, target, cwd=None):
```
misl6, line 479: Seems we still have a reference to `file_rename` here:
https://github.com/kivy/buildozer/blob/b1f0e61ef30462df021f08ef1dc15bf10e40e8d6/buildozer/targets/ios.py#L313

misl6 (changes requested):
From a quick check, I've found (and added a comment) one reference of `file_rename` which needs to be migrated, can you take care of it?

Julian-O:
@misl6:

Good spot. I do remember seeing the log message directly above it ('Moving IPA to bin...' sounds like it is throwing away some beer) so I don't know how I missed this one. Fixed.

[I also accidentally gave a sneak preview of the *next* PR by committing in the wrong branch. I've undone that too.]

misl6:
> ('Moving IPA to bin...' sounds like it is throwing away some beer)

😂

misl6 (approved):
LGTM. Thank you!

PR #1670, opened 2023-08-26
Use buildops for file_exists and file_matches

Julian-O:
This is part of a larger refactor, to reduce the size and complexity of the Buildozer class.

* Remove `file_matches()` method from the Buildozer class.
* Change the singular reference to it in `android.py` to use the buildops function.
   * This raises the question: Will this be useful to other targets in the future, or should this be moved into the android target? Left for future refactors.

* Remove the `file_exists()` method from the Buildozer class.
* Change all the references to use the buildops function.
   * Note for consistency with the other buildop functions, buildops.file_exists() takes a single parameter rather than `*args` and doesn't join the path together. One client call needed to be changed to accommodate that.
   * Note that in several places, the code checks for the existence of *directories* with os.path.exists(). I did **not** change these over because:
     * Historically, the name has been **file**_exists().
     * With the join statement removed, this function doesn't add anything to calling `os.path.exists()` or `pathlib.Path.exists()`, so enforcing it on the code seemed counter-productive. I went with for consistency with existing code.
 I can easily be convinced to change this.
 
* Update mocks

### DIFF
--- a/buildozer/__init__.py
+++ b/buildozer/__init__.py
@@ -66,7 +66,7 @@ class Buildozer:
 
         self.logger = Logger()
 
-        if exists(filename):
+        if buildops.file_exists(filename):
             self.config.read(filename, "utf-8")
             self.check_configuration_tokens()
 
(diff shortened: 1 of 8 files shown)


misl6 (approved):
LGTM. Thank you!

PR #1671, opened 2023-08-27
Move checkbin and download to use buildops.

Julian-O:
This is part of a larger refactoring to simplify the Buildozer class.

* Remove checkbin() and download() methods (plus some definitions only used by download().
* Migrates all references to use the Buildops equivalents.
* Migrate shelling out to "curl"  to use download instead.

Note: The Buildops versions have completely different implementations to the original Buildozer ones, but the semantics are the same. New versions are more platform-independent, expected to be faster and don't use deprecated  library calls.

[Note to reviewers: This should be a straightforward one, but heads up that I am expecting the *next* PR to be a doozy.

### DIFF
--- a/buildozer/targets/osx.py
+++ b/buildozer/targets/osx.py
@@ -47,8 +47,7 @@ class TargetOSX(Target):
                 self.logger.info('Downloading kivy...')
                 try:
                     buildops.download(
-                        f'https://kivy.org/downloads/'
-                        f'{current_kivy_vers}/Kivy.dmg',
+                        f'https://kivy.org/downloads/{current_kivy_vers}/Kivy.dmg',
                         'Kivy.dmg',
                         cwd=cwd
                     )
--- a/buildozer/targets/osx.py
+++ b/buildozer/targets/osx.py
@@ -8,6 +8,7 @@ if sys.platform != 'darwin':
 
 from os.path import exists, join, abspath, dirname
 from subprocess import check_call, check_output
+import urllib
 
 import buildozer.buildops as buildops
 from buildozer.target import Target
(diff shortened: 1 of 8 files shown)


buildozer/targets/osx.py
```diff
@@ -44,16 +45,19 @@ def download_kivy(self, cwd):
...
-
-                if status_code == "404":
+                try:
+                    buildops.download(
+                        f'https://kivy.org/downloads/'
+                        f'{current_kivy_vers}/Kivy.dmg',
+                        'Kivy.dmg',
+                        cwd=cwd
+                    )
+                except urllib.Error:
```
misl6, line 55: Does `urllib.Error` exists?

These: https://github.com/python/cpython/blob/3.11/Lib/urllib/error.py are the valid errors that `urllib` can raise.

buildozer/targets/osx.py
```diff
@@ -44,16 +45,19 @@ def download_kivy(self, cwd):
...
-                    'curl', '-L', '--write-out', '%{http_code}',
-                    '-o', 'Kivy.dmg',
-                    f'https://kivy.org/downloads/{current_kivy_vers}/Kivy.dmg'),
-                    cwd=cwd)
-
-                if status_code == "404":
+                try:
+                    buildops.download(
+                        f'https://kivy.org/downloads/'
+                        f'{current_kivy_vers}/Kivy.dmg',
```
misl6, line 51: Minor: Can we improve the readability by avoiding the string slicing?

misl6 (changes requested):
HI @Julian-O !

I've found an issue + a potential improvement, can you take care of it?

buildozer/targets/osx.py
Julian-O: Will fix. This one annoys me because I deliberately set up an prototype to see what errors actually got raised, and then I managed to write it down wrong. And static code analysis didn't save me. :-(

buildozer/targets/osx.py
Julian-O: The full string went over the 79/80 character line limit, so I split it in two.

Would putting them in parenthesis and adding them together help?

I will go back to breaking the line length rules for now, so this minor issue isn't blocking us.

Julian-O:
Changes made.

Test failed due to:

> [Integration (apple-silicon-m1, 3.9.7)](https://github.com/kivy/buildozer/actions/runs/5990305371/job/16247484457)
The self-hosted runner: apple-silicon-humongous-avocet lost communication with the server. Verify the machine is running and has a healthy network connection. Anything in your workflow that terminates the runner process, starves it for CPU/Memory, or blocks its network access can cause this error.

Doesn't sound like me. Could you please trigger it again?

misl6 (approved):
LGTM. Thank you!

(re-triggered the CI jobs, and let's wait a completion, but I do not expect a failure due to your changes)

PR #1672, opened 2023-08-27
Fix a typo introduced during PR #1669

misl6:
CC: @Julian-O 

We both did not see that the method name was wrong, since we likely were too much focused on `self.buildozer` --> `buildops`.

This only means that we need to improve our tests.

### DIFF
--- a/buildozer/targets/ios.py
+++ b/buildozer/targets/ios.py
@@ -311,7 +311,7 @@ class TargetIos(Target):
             cwd=build_dir)
 
         self.logger.info('Moving IPA to bin...')
-        buildops.file_rename(ipa_tmp, ipa)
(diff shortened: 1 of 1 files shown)


Julian-O:
@misl6: Yeah, I noticed this one, and had included a fix in the next PR.

Gotta agree with you about needing better test coverage, but that is a tough ask. Ironically, it was a static code analysis from my IDE that first brought it to my attention. But `ios.py` has 12 warnings from my IDE, and `buildozer\__init__.py` has over 40, so it is easy to miss new stuff. As a result, I was left pondering how and when to submit edits that just reduced warning counts. [And when to slip "black" into the commit hooks.]

PR #1682, opened 2023-09-08
Rewrite section in documentation about on source include/excludes

Julian-O:
The section in the documentation on source.includes/excludes is:

* Incomplete. It doesn't mention `include_patterns` or "hidden" files.
* Wrong: It says "If [`source.exclude_exts` is] 
  empty, no files will be excluded based on their extensions." Reading the code reveals that if `source.exclude_exts` is empty, it has no effect.
* Unclear: Little mention on how the different options interact.

This is a re-write to attempt to fix that.


misl6 (approved):
LGTM. Thank you!

PR #1683, opened 2023-09-08
Add tests of the documentation.

Julian-O:
The RST files in a pull request might be malformed. URLs might be broken. These should be included in the unit-tests so they are immediately noticed.

This runs a Sphinx build on the documentation (and throws away the result). It also runs a URL check.

Note: The existing documents have some warnings - look for a separate PR to address those.

I did experiment, and confirm that an error like a bad URL causes this test to fail.


misl6 (approved):
Nice idea! We should add the same elsewhere!

LGTM. Thank you!

PR #1684, opened 2023-09-08
Correct documentation configuration

Julian-O:
There is an internal link to 'specifications' that requires a Sphinx extension to be installed to work.

There are no  custom static files, so removed spurious mention to `_static` directory,


misl6 (approved):
LGTM. Thank you!

PR #1687, opened 2023-09-10
Now Github Actions provides `python3` via `setup-python` also for Apple Silicon macs

misl6:
Finally, no more apple-silicon specific workarounds! 🎉 (at least on selected Python3 versions, but that's enough for `kivy-ios`)

Similar PR on `kivy-ios`: https://github.com/kivy/kivy-ios/pull/829


Julian-O (approved):
I ain't an official reviewer, but looks good to me.

PR #1688, opened 2023-09-10
Use suggested (latest) Python3 version to perform Android builds

misl6:
`buildozer` is already tested against all the supported Python versions.

During Android builds there's no need to stick on a specific Python version.


Julian-O (approved):
I went off to research whether YAML needs single-quotes around `'3.x'`. The answer seems to be no, but it doesn't hurt *and* YAML's string syntax is a horrifying nightmare.

So, LGTM.

PR #1689, opened 2023-09-10
Improve Buildozer's venv handling.

Julian-O:
* The mechanism used to determine if it was currently running in a venv was non-standard. Used technique recommended by [venv documentation](https://docs.python.org/3/library/venv.html#how-venvs-work).
* Rather than call a subprocess to run venv, use the built-in venv library.
* Rather than check if an attribute exists, define it in `__init__()` and see if it has changed.
* `self.venv` contained a directory path, but the folder name didn't need to be stored after the method returned. A Boolean was all that was required.

Also, some trivial clean ups thrown in:
* Comment improvements.
* Copy command didn't need to be logged; buildops will do that.
* Directory's existence doesn't need to be checked; buildops will do that.

### DIFF
--- a/buildozer/__init__.py
+++ b/buildozer/__init__.py
@@ -18,6 +18,7 @@ import sys
 from sys import exit
 import textwrap
 import warnings
+import venv
 
 import buildozer.buildops as buildops
 from buildozer.jsonstore import JsonStore
@@ -38,6 +39,7 @@ class Buildozer:
         self.state = None
         self.build_id = None
         self.config = SpecParser()
+        self._venv_created = False
 
         self.logger = Logger()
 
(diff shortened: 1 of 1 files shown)


buildozer/__init__.py
```diff
@@ -225,19 +227,25 @@ def check_application_requirements(self):
...
+
+        # TODO: Make more general - filter at the first non [a-z0-9_-] char?
         onlyname = lambda x: x.split('==')[0]  # noqa: E731
+
         requirements = [x for x in requirements if onlyname(x) not in
                         target_available_packages]
 
-        if requirements and hasattr(sys, 'real_prefix'):
+        # Technique defined in venv library documentation.
+        currently_in_venv = sys.prefix != sys.base_prefix
```
misl6, line 238:
```suggestion
        # Technique defined in venv library documentation.
        # See: https://docs.python.org/3/library/venv.html#how-venvs-work
        currently_in_venv = sys.prefix != sys.base_prefix
```

Minor suggestion

Julian-O:
Good idea. Updated.

misl6 (approved):
LGTM. Thank you!

PR #1690, opened 2023-09-10
Clean up inheritance of targets

Julian-O:
1) Target class contained Android-only attribute "artifact_format". Moved that code into TargetAndroid class.

2) TargetIos had a `check_build_prepared()` method that was ineffective.

`prepare_for_build()` sets `target._build_prepared` to True. Then `check_build_prepared()` inexplicably set it to False. Then `build()` checks for its existence, rather than its value [this is code smell that is on my list for later], so the fact that it had been changed made no difference.

So, I removed the method, and the call to it.

2) TargetOSX had lots of methods that were identical to its base class implementations (three more, once the above steps were done). They don't need to be specified; they are inherited. I removed them. `osx.py` is now 20% smaller.

### DIFF
--- a/buildozer/target.py
+++ b/buildozer/target.py
@@ -15,7 +15,6 @@ class Target:
     def __init__(self, buildozer):
         self.buildozer = buildozer
         self.build_mode = 'debug'
-        self.artifact_format = 'apk'
         self.platform_update = False
         self.logger = Logger()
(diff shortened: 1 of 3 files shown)


misl6 (approved):
Nice!

LGTM. Thank you!

PR #1692, opened 2023-09-10
Add alternate unzip implementation

Julian-O:
Turns out Linux and the `zipfile` package don't get along, while Windows and `unzip` don't get along.

Provide both implementations, so it always works.

[This doesn't meet my stated goal of making all Windows-specific improvements also include making the code cleaner for other developers too. Sorry.]

### DIFF
--- a/buildozer/buildops.py
+++ b/buildozer/buildops.py
@@ -22,6 +22,7 @@ import time
 import tarfile
 from threading import Thread
 from urllib.request import Request, urlopen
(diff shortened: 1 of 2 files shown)


misl6 (approved):
> This doesn't meet my stated goal of making all Windows-specific improvements also include making the code cleaner for other developers too. Sorry.

I think that we all can live with it in order to have better support on Windows. 😀

PR #1694, opened 2023-09-11
Remove some attributes defined outside of __init__()

Julian-O:
Attributes of a class instance should, where possible, be defined in the `__init__()` method, for clarity (especially about lifetimes). Pylint has a [warning about this](https://pylint.readthedocs.io/en/latest/user_guide/messages/warning/attribute-defined-outside-init.html), but I can't see one from Flake8.

TargetAndroid has an attribute `_serials` which is defined  (*sometimes*) outside of `__init()__`, and its value is checked with `hasattr()`. Changed it to act like a normal instance attribute, defined in `__init__()`.

Buildozer() goes even further. It monkey-patches two attributes, `_build_prepared` and `_build_done` onto its associated target. These attributes have the same lifetime as the Buildozer instance. I removed the monkey-patching, and made them normal instance attributes of Buildozer, defined in `__init__()`.

### DIFF
--- a/buildozer/__init__.py
+++ b/buildozer/__init__.py
@@ -40,6 +40,8 @@ class Buildozer:
         self.build_id = None
         self.config = SpecParser()
         self._venv_created = False
+        self._build_prepared = False
+        self._build_done = False
 
         self.logger = Logger()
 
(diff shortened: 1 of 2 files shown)


misl6 (approved):
LGTM. Thank you!

PR #1696, opened 2023-09-13
Remove empty except clauses

Julian-O:
[Empty except clauses are code smell](https://stackoverflow.com/questions/14797375/should-i-always-specify-an-exception-type-in-except-statements/). At the very least, they can unintentionally catch KeyboardInterrupt and SystemExit exceptions. At worse, they can hide bugs.

Looking through Buildozer code, there are four empty except clauses in `android.py`.

One protects a call to set_win32_java_home() (on Windows machines), Given win32 isn't yet supported, this is moot, but there doesn't seem to be a justification for ignoring exceptions in this scenario - if setting the home fails, Buildozer should fail. Deleted try block.

One protects a parse of the app version number. The called code already catches the most obvious exception (a badly-formatted version number). There don't seem to be any other expected exceptions, so there is not reason to swallow this exception. Deleted try block.

One is harder to understand. After calling `execute_built_package()`, a try block attempts to call it **again**, surrounded by calls to run a pre-build and post-build hook on buildozer. However, the hook is never defined anywhere. This always dies with an `AttributeError` which is swallowed and ignore by the empty exception block, without the duplicate call to build. I am guessing it is a carry-over from an attempt to make Buildozer support hooks. This is an example of why not to use empty Exception blocks! Removed the try block and the protected code.

The fourth example protects an attempt to decode Unicode. `content` is read from a IO object opened in 'r' mode, so it will always be a `str` and will never have a `decode()` attribute, so this will always fail. Removed the code in the try block and just kept the except: part.

### DIFF
--- a/buildozer/targets/android.py
+++ b/buildozer/targets/android.py
@@ -31,7 +31,6 @@ import shlex
 from shutil import which
 from sys import platform, executable
 from time import sleep
-import traceback
 
 from distutils.version import LooseVersion
 import pexpect
@@ -273,10 +272,7 @@ class TargetAndroid(Target):
 
     def check_requirements(self):
         if platform in ('win32', 'cygwin'):
-            try:
-                self._set_win32_java_home()
-            except:
-                traceback.print_exc()
+            self._set_win32_java_home()
             self.adb_executable = join(self.android_sdk_dir, 'platform-tools',
                                 'adb.exe')
             self.javac_cmd = self._locate_java('javac.exe')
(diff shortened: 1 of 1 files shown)


AndreMiras:
Hey @Julian-O I agree on that one, not sure how I didn't review it before.
Let's indeed clean them up and if it introduces regression then let's have them more explicitly handled and unit tested.
Would you like to rebase and fix conflicts?

Julian-O:
Looks like all of the instances of this have been fixed independently in the meantime. Closing.

PR #1709, opened 2023-09-28
Make the logger utilize rich to make the logs look nicer.

matan-h:
This would make the logs nicer by using the awesome [`rich`](https://rich.readthedocs.io/en/stable/logging.html) library.
The current logs can be long and using `rich` will make them more readable and easier to find specific messages

### DIFF
--- a/buildozer/logger.py
+++ b/buildozer/logger.py
@@ -3,48 +3,21 @@ Logger
 ======
 
 Logger implementation used by Buildozer.
+The log itself is performed using rich and using the python logging framework.
 
-Supports colored output, where available.
+rich automatically Supports colored output, where available.
 """
 
-from os import environ
+import logging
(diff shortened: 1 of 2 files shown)


Julian-O:
Some background:

* Kivy, Buildozer and python-for-android each have their own custom logging solutions, that don't match!
* I spent some effort a year ago to make Kivy logging follow the standard `logging` infrastructure.

I think you should break this into two steps.

The first is to change Buildozer's custom logger system to use the conventional python's `logging` class. I would like to see the Bulldozer logger inherit from Python `logging.logger`, rather than aggregating/shimming it.

After that, substituting in the new Rich handler should be trivial - I am not sure if it will be appreciated, as it won't match Kivy or Python-For-Android in appearance, but it should be trivial. (`logging` has a config system that might allow the user to choose.)

PR #1719, opened 2023-10-12
fix for OSX related builds issues

butuzov:
This PR intends to fix some of the issues related to OSX builds

1. Downloading proper archives and dmgs  
2. Copying whole fodler of mac app   
3. Adding requirements for osx builder (its goign to fail without it)

### DIFF
--- a/buildozer/targets/osx.py
+++ b/buildozer/targets/osx.py
@@ -62,9 +62,9 @@ class TargetOSX(Target):
                     sys.exit(1)
 
             self.logger.info('Extracting and installing Kivy...')
-            check_call(('hdiutil', 'attach', cwd + '/Kivy.dmg'))
-            buildops.file_copy(
-                '/Volumes/Kivy/Kivy.app', './Kivy.app', cwd=cwd)
+            check_call(('hdiutil', 'attach', f'{cwd}/Kivy.dmg'))
+            buildops.file_copytree(
+                '/Volumes/Kivy/Kivy.app', f'{cwd}/Kivy.app')
 
     def ensure_kivyapp(self):
         self.logger.info('check if Kivy.app exists in local dir')
--- a/buildozer/targets/osx.py
+++ b/buildozer/targets/osx.py
@@ -3,12 +3,14 @@ OSX target, based on kivy-sdk-packager
 """
 
 import sys
+
 if sys.platform != 'darwin':
     raise NotImplementedError('This will only work on osx')
 
-from os.path import exists, join, abspath, dirname
-from subprocess import check_call, check_output
 import urllib.error
+from os.path import abspath, dirname
+from pathlib import Path
+from subprocess import check_call, check_output
 
 import buildozer.buildops as buildops
 from buildozer.target import Target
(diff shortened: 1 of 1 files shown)


buildozer/targets/osx.py
```diff
@@ -62,9 +62,9 @@ def download_kivy(self, cwd):
                     sys.exit(1)
 
             self.logger.info('Extracting and installing Kivy...')
-            check_call(('hdiutil', 'attach', cwd + '/Kivy.dmg'))
-            buildops.file_copy(
-                '/Volumes/Kivy/Kivy.app', './Kivy.app', cwd=cwd)
+            check_call(('hdiutil', 'attach', f'{cwd}/Kivy.dmg'))
+            buildops.file_copytree(
+                '/Volumes/Kivy/Kivy.app', f'{cwd}/Kivy.app')
```
misl6, line 67: What about using `pathlib.Path` here, instead of the f-string?

misl6 (changes requested):
Hi @butuzov !

Ugh. That issue would be not there if we had a CI job for osx.

I've added a suggestion, can you please share your thoughts about it?

buildozer/targets/osx.py
butuzov: Up to you! TBH, there like 4 different methods of joining pathes (string concatenation, `dot-foramt`, `os.path.join` (even if we work in only darwin env), fstrings). I would maybe stent some time in future just switching to `pathlib.Path` concatenation somewhere in future, not as part of this small PR (event with that I can't make it publishing for mac, but this issue  is onto osx builder package). lets make small steps?

cederom:
+1 :-)

buildozer/targets/osx.py
Julian-O: @butuzov: Yes, indeed. There have already been quite a few small steps towards `pathlib` in Buildozer, with more in the pipeline (eventually). So, sticking with `pathlib` here will help us not take any steps backwards. Thanks!

butuzov:
Got it. sorry for delay, will try to fix it asap.

butuzov:
While this still WIP, it would be nice if you can check the code, and perform code review. Its mostly ready, how ever i need to confirm issues in Kivy SDK & in `osx.py` before merging.


P.S.
I am sorry for my slow speeed, my conditions of work isn't best.

buildozer/targets/osx.py
Julian-O, line 1: re: [buildozer/targets/osx.py](https://github.com/kivy/buildozer/pull/1719/commits/b9e8f021ce676a5eb210acd676b7287370002bf4#diff-323c235975e316e212a3c3c3467612dc97e4bc47548c599fa8eab4e0116d8a21)

The first line of `buildops.download()` is 

     url = url + str(filename)
     
 [I believe I did it that way to work most similarly to the old code, even though it is an odd choice.]
     
So it shouldn't be necessary to add the filename. Was this working before the change?

buildozer/targets/osx.py
```diff
@@ -62,9 +62,9 @@ def download_kivy(self, cwd):
                     sys.exit(1)
 
             self.logger.info('Extracting and installing Kivy...')
-            check_call(('hdiutil', 'attach', cwd + '/Kivy.dmg'))
-            buildops.file_copy(
-                '/Volumes/Kivy/Kivy.app', './Kivy.app', cwd=cwd)
+            check_call(('hdiutil', 'attach', f'{cwd}/Kivy.dmg'))
```
Julian-O: It would be nice if this used PathLib paths, but I acknowledge this will only run on OSX so this is purely aesthetic, not important.

buildozer/targets/osx.py
```diff
@@ -93,19 +105,21 @@ def build_package(self):
         author = bc.getdefault('app', 'author', '')
 
         self.logger.info('Create {}.app'.format(package_name))
-        cwd = join(self.buildozer.platform_dir, 'kivy-sdk-packager-master', 'osx')
-        osx_deps = open(join(cwd, 'requirements.txt')).read()
+        osx_deps = ( self.kivy_app_dir / "requirements.txt" ).read_text()
```
Julian-O, line 108: Huh! `read_text()` New one on me. Thanks!

buildozer/targets/osx.py
```diff
@@ -94,6 +94,8 @@ def build_package(self):
 
         self.logger.info('Create {}.app'.format(package_name))
         cwd = join(self.buildozer.platform_dir, 'kivy-sdk-packager-master', 'osx')
+        osx_deps = open(join(cwd, 'requirements.txt')).read()
```
Julian-O: I know this is OSX and so it doesn't probably matter, but this is an anti-pattern for file reading, because the file won't be closed until the GC notices, and on some OSes this means the file is unavailable for other processes.

buildozer/targets/osx.py
```diff
@@ -94,6 +94,8 @@ def build_package(self):
...
+        app_deps += osx_deps
         # remove kivy from app_deps
         app_deps = [a for a in app_deps.split('\n') if not a.startswith('#') and a not in ['kivy', '']]
```
Julian-O: Question: requirements.txt files can contain quote complex entries, specifying versions based on platform, Python versions etc.

Are these parsed by app_deps?

Julian-O (commented):
I am not an official reviewer - please treat my comments as suggestions.

butuzov:
> I am not an official reviewer - please treat my comments as suggestions.

First, of all. I am sorry for slow responses. I am not in working (regular office job conditions now). Secondly, thank you for your comments. Will respond in a second.

buildozer/targets/osx.py
butuzov: Not really sure, why this comment realted to line 400. FIrst inital commit of this PR was fixing issue with kivy dmg copy, that kinda relates to this code. Should this be skipped?

buildozer/targets/osx.py
butuzov: It using now, I think you reviewed outdated code. At least this line is.

buildozer/targets/osx.py
butuzov: and it's `contexted`!

buildozer/targets/osx.py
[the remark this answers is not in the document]
butuzov: Was there? I am so new to Kivy, that I might miss it.  However without this code, you can't move next. YOu can't create `.app` without `docopt` and `sh` if you didn't hae them before.  (its comming from `kivy-sdk-packager-master`) and btw, this code also outdated.

butuzov: Yeah, my bad. Anyway, this already fixed to `.read_text()`, witch closes descriptor.

buildozer/targets/osx.py
butuzov: We better ask @akshayaurora (code is from 5 years back in time). But as you see parsing is quite simple...

buildozer/targets/osx.py
Julian-O: Oh, this is what I think may have happened:

You created a branch OSX-related fixes.
You committed a fix, and I assume raised a PR.
Then you committed a different fix for a different issue in the same branch, and raised a PR.
You repeated this several times.

When you do this, each PR includes *all of your previous changes*, and puts them all up for review. I was making comments on earlier commits you made.

If you create a different branch for each commit, that doesn't happen (although it means you don't get to rely on changes you have already made - unless you wait for them to be merged first)

[I hope I am right about what you did, or I just made this far more confusing!]

butuzov: Yeah, almost. 
- This initial PR changed 2 related to buildops things + installs requirements.txt (as it failed without it).
- I might missunderstand initial request, but it looks like i was asked to switch to `pathlib.Path`, at elast for `osx.py`, which i did.

akshayaurora:
This issue is already solved in master.

PR #1755, opened 2023-12-10
Stop No Response action running when unneeded.

Julian-O:
1) The No Response action is running every 5 minutes, wasting CPU. Once a day is plenty.

2) If you fork the repository in order to contribute, and you want the CI test actions to run on each commit, you can inherit the original actions from the main repo. However, that also inherits No Response action, which can't run without an issue tracker. As a result, you receive a spurious error email every time No Response runs. While it is possible to individually exclude actions with some sort of regular expression parameter in the GitHub web-site, it is neater to have the action automatically skip if it isn't in the Kivy/* domain.

### DIFF
--- a/.github/workflows/no-response.yml
+++ b/.github/workflows/no-response.yml
@@ -6,11 +6,13 @@ on:
   issue_comment:
     types: [created]
   schedule:
-    # Schedule for five minutes after the hour, every hour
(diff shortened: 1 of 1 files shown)


misl6 (approved):
LGTM. Thank you!

PR #1756, opened 2023-12-10
Update installation.rst to add support for Debian 12

AJRepo:
This adds documentation support for Debian 12. 

It also adds a bash line break to the Ubuntu commands to keep the text from being offscreen to the right, requiring scrolling to see.

### DIFF
--- a/docs/source/installation.rst
+++ b/docs/source/installation.rst
@@ -22,6 +22,42 @@ If you would like to install the latest version still under development::
 Targeting Android
 -----------------
 
+Android on Debian 12 (Bookworm)
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The below Installation and Build method uses a virtual python environment. 
+
+.. note::
+
+Debian Notes: 
+
+* `libncurses-dev` replaces `libncurses5-dev` and `libncursesw5-dev` (older OS versions)
+
+* `openjdk-17-jdk-headless`: Fixes the error "not found javac" 
+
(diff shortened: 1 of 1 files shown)


docs/source/installation.rst
```diff
@@ -22,6 +22,42 @@ If you would like to install the latest version still under development::
...
+
+* `python3-sdl2`: Fixes errors like "clang ... error no input file ... " 
+
+* If in running `buildozer` you get *not found* errors from a missing package outside the python virtual env then delete the python virtual environment, add the missing package (apt install ...), and re-create the python virtual environment. 
+
+Additional installation required to support Android::
+
+    sudo apt update
+    sudo apt install git zip unzip openjdk-17-jdk-headless python3-pip autoconf libtool pkg-config zlib1g-dev \
+        python3-kivy autoconf-archive gnu-standards gettext python3-venv libffi-dev python3-sdl2 libncurses-dev libssl-dev
```
misl6, line 46: If `python3-kivy` and `python3-sdl2` are needed, then something is wrong, as the host environment should not interfere with the cross-compilation process in that case.

Can you please post the full traceback?

AJRepo: I'm afraid that the system on which I did this has already been rebooted so I no longer have the traceback for the times it didn't work. So we'd have to create a fresh Debian 12 VM or container to re-test without the host packages you are wondering about. I did have to add them to get past those compilation errors. My understanding of the python virtualenv is that it will look in the host env also, so you either have to add stuff in the host env or  the virtualenv.

PR #1761, opened 2024-01-05
Add macOS build instructions, upgrades

jooray:
Add macOS build instructions (requires different build type on Apple Silicon because it needs x86-amd64 emulation)
Upgrade Ubuntu to latest LTS
Upgrade JDK to required JDK 17, builds did not work with old JDK


akshayaurora (approved):

akshayaurora:
@jooray thanks!

PR #1765, opened 2024-01-30
Fix download url for kivy-sdk-packager osx.py

akshayaurora:


akshayaurora:
@Julian-O We should probably have a test....

PR #1768, opened 2024-01-30
Fix Osx Target to properly use buildozer.buildops

akshayaurora:
OSX Target was incorrectly using buildozer.buildops


Julian-O (approved):
I am still concerned there is no integration test that would have captured these problems.

PR #1804, opened 2024-05-04
Use `ubuntu:22.04` instead of `ubuntu:latest` in `Dockerfile`

manatlan:
Use ubuntu:22.04 not ubuntu LTS, because the container is not ready for ubu 24.04 ;-)


misl6 (approved):
LGTM. Thank you!

PR #1839, opened 2024-07-17
Set Android min API to 24

Cheaterman:
Avoids a very common error:
```
ImportError: cannot import name 'if_indextoname' from 'socket'
```


T-Dynamos:
@Cheaterman https://github.com/kivy/python-for-android/pull/3007#discussion_r1592817369

Cheaterman:
I see, but practicality beats purity, and this has been a thorn in every new user's side for way too long. @misl6 Are there viable alternatives? If not, we should probably bite the bullet and merge this.

PR #1879, opened 2025-01-05
Explain Python version restrictions in README.md

flosincapite:
I note that there is already [a PR to support Ubuntu 24](https://github.com/kivy/buildozer/pull/1824), but hopefully this note will help guide others that run into issues in the meantime.


AndreMiras:
Thanks for the pull request. I'm not sure in which case this is happening actually.

PR #1899, opened 2025-03-17
Android sdkmanager proxy config

broccoliboy:
This PR configures android sdkmanager to use system proxy info from env var http_proxy, which is needed for buildozer android to work on corporate networks with a proxy in place. Related to #142 but this seems like a simpler implementation that leverages env vars that are typically already defined while working in a corporate proxy setting.

### DIFF
--- a/buildozer/targets/android.py
+++ b/buildozer/targets/android.py
@@ -205,11 +205,9 @@ class TargetAndroid(Target):
         # Sdkmanager only supports proxy types `http` and `socks`, so even if proxy url is https,
         # only pass host and port with `http` type.
         for key in ['HTTP_PROXY', 'http_proxy', 'HTTPS_PROXY', 'https_proxy']:
-            proxy = os.environ.get(key)
             try:
-                host, port = proxy.split(':')[-2:]
-                host = host.strip('/')
-                command.extend(['--proxy=http', f'--proxy_host={host}', f'--proxy_port={port}'])
+                host, port = os.environ.get(key).split(':')[-2:]
+                command.extend(['--proxy=http', f'--proxy_host={host.strip("/")}', f'--proxy_port={port}'])
                 break
             except:
                 pass


buildozer/targets/android.py
```diff
@@ -199,6 +199,21 @@ def _sdkmanager(self, *args, **kwargs):
...
         kwargs['cwd'] = kwargs.get('cwd', android_sdk_dir)
         command = [self.sdkmanager_path, f"--sdk_root={android_sdk_dir}", *args]
+        
+        # Attempt to pass proxy settings to sdkmanager command using the --proxy, --proxy_host,
+        # and --proxy_port arguments by checking common environment variables for proxy settings.
+        # Sdkmanager only supports proxy types `http` and `socks`, so even if proxy url is https,
+        # only pass host and port with `http` type.
+        for key in ['HTTP_PROXY', 'http_proxy', 'HTTPS_PROXY', 'https_proxy']:
+            proxy = os.environ.get(key)
+            if proxy:
```
kuzeyron: If statement could be removed and just have try/except there

buildozer/targets/android.py
```diff
@@ -200,6 +200,20 @@ def _sdkmanager(self, *args, **kwargs):
...
         command = [self.sdkmanager_path, f"--sdk_root={android_sdk_dir}", *args]
 
+        # Attempt to pass proxy settings to sdkmanager command using the --proxy, --proxy_host,
+        # and --proxy_port arguments by checking common environment variables for proxy settings.
+        # Sdkmanager only supports proxy types `http` and `socks`, so even if proxy url is https,
+        # only pass host and port with `http` type.
+        for key in ['HTTP_PROXY', 'http_proxy', 'HTTPS_PROXY', 'https_proxy']:
+            proxy = os.environ.get(key)
+            try:
+                host, port = proxy.split(':')[-2:]
```
kuzeyron: `proxy = os.environ.get(key)` could be set directly with `host, port = proxy.split(':')[-2:]` so it becomes `host, port = os.environ.get(key).split(':')[-2:]`

kuzeyron (approved):

PR #1958, opened 2025-09-06
Update Python workflow for version and action upgrades

cclauss:
* https://github.com/actions/runner-images?tab=readme-ov-file#available-images `macos-latest` is macOS 15 ARM64
* https://github.com/actions/setup-python/releases
* https://github.com/actions/checkout/releases
* https://www.python.org/download/pre-releases
* https://www.python.org/downloads/release/python-3140rc2


AndreMiras (approved):
LGTM thanks

PR #1969, opened 2025-10-01
Upgrade from deprecated macos-13 to macos-15-intel in CI

cclauss:
Upgrade from deprecated `macos-13` GitHub Actions runner to `macos-15-intel`.

The currently available GitHub Actions macOS runners are:
| macOS Version | runner.arch |
|---------------|-------------|
| macos-13 | X64 |
| macos-14 | ARM64 |
| macos-15 | ARM64 |
| macos-15-intel | X64 |
| macos-26 | ARM64 |
| macos-latest | ARM64 |
* Let's prepare for actions/runner-images#13046
* Like: kivy/kivy#9162

Should a set of test runs be made on `macos-26`?


.github/workflows/ios.yml
```diff
@@ -7,10 +7,10 @@ jobs:
     strategy:
       matrix:
         # macos-latest (ATM macos-15) runs on Apple Silicon,
-        # macos-13 runs on Intel
-        runs_on: [macos-latest, macos-13]
+        # macos-15-intel runs on Intel
+        runs_on: [macos-latest, macos-15-intel]
```
AndreMiras, line 11: How about using explicit naming for both so we know what we're talking about?

.github/workflows/android.yml
```diff
@@ -28,14 +28,14 @@ jobs:
...
-    - name: SDK, NDK and p4a download
+    - name: SDK, NDK, and p4a download
       run: |
         sed -i.bak "s/# android.accept_sdk_license = False/android.accept_sdk_license = True/" buildozer.spec
         sed -i.bak "s/#p4a.branch = master/p4a.branch = develop/" buildozer.spec
         buildozer android p4a -- --help
     # Install OS specific dependencies
     - name: Install Linux dependencies
-      if: matrix.os == 'ubuntu-latest'
+      if: runner.os == 'Linux'
```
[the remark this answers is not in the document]
cclauss: Yes positive.  The ___detecting the operating system___ link in the commit message.

.github/workflows/ios.yml
AndreMiras: I mean, why not `macos-15` rather than `macos-latest` so we're explicit about what version we're trying to run.
also any plan on running `macos-14`?
Also I like that the label `macos-15-intel` is already explicit about the architecture, that way we can remove the comment above it

PR #1983, opened 2025-11-11
Update __init__.py to fix #1982

### LINKED ISSUE #1982: Buildozer complains about missing although actually existing and readable/writable/executable main.py in the app directory.
After long investigation as to why buildozer complains about a missing main.py file in my app directory (BUILD FAILURE: No main.py(c) found in your app directory.), although main.py exists in my app directory, is readable, writable and can be run with python, is included in buildozer.spec (even explicitly included, doesn't work), I found the reason and think it's worth at least sharing this for others that could be in the same case, because it's very hard to diagnose.

In __init__.py of buildozer, inside _copy_application_sources(self), these lines:

        for root, dirs, files in walk(source_dir, followlinks=True):
            # avoid hidden directory
            if True in [x.startswith('.') for x in root.split(sep)]:
                continue

seem to be intended to ignore files that live inside a hidden subfolder.

But, if any of the parents of the source_dir folder is hidden itself, then all the project's content is ignored, which results in the BUILD FAILURE mentioned above. (Actually, no file at all is copied to the build directory, hence no main.py).

I'd suggest either this fix:

        for root, dirs, files in walk(source_dir, followlinks=True):
            # avoid hidden directory
            rel_path = root[len(source_dir):].lstrip(sep)
            if rel_path and True in [x.startswith('.') for x in rel_path.split(sep)]:
                continue

which would still ignore hidden subdirectories, but not ignore the project itself if one of its parents is hidden; or alternatively, if this behavior is actually intended, issue a warning that a project which has a hidden parent cannot be built?

nicolashainaux:
Problem is described here: #1982

### DIFF
--- a/tests/test_buildozer.py
+++ b/tests/test_buildozer.py
@@ -303,12 +303,6 @@ class TestCopyApplicationSources(unittest.TestCase):
             f.write(content)
         return filepath
 
-    # def create_dir(self, relpath):
-    #     """Helper to create a directory in source_dir."""
-    #     dirpath = os.path.join(self.source_dir, relpath)
-    #     os.makedirs(dirpath, exist_ok=True)
-    #     return dirpath
-
     @mock.patch('buildozer.buildops')
     def test_ignore_hidden_files_and_directories_in_source(self,
                                                            mock_buildops):


AndreMiras (dismissed):
Looking good, thanks for investigating and fixing.
Having tests would be great too, do you want to give a stab at it?

nicolashainaux:
OK, I need a few days and give you a proposal.

nicolashainaux:
Finally, as your tests didn't seem to cover Buidlozer._copy_application_sources at all, I have included tests to cover more of its code (not only the hidden parent case), though I am not 100% sure it covers all cases.

In order to make changes in a cleaner way, I have gathered the tests of in a new dedicated class TestCopyApplicationSources(unittest.TestCase) with its own setUp, tearDown and helpers. So I think it's clean, clear and does not interfere with existing tests.

It looks good, on my local linux I get this (I only check python 3.11 here):

```
(buildozer-py311) buildozer [patch-1 ↑·1|✔] $ pytest -vv tests/test_buildozer.py
================================================================================== test session starts ==================================================================================
platform linux -- Python 3.11.13, pytest-9.0.1, pluggy-1.6.0 -- /home/nico/dev/.venv/buildozer-py311/bin/python
cachedir: .pytest_cache
rootdir: /home/nico/dev/buildozer
collected 14 items                                                                                                                                                                      

tests/test_buildozer.py::TestBuildozer::test_android_ant_path PASSED                                                                                                              [  7%]
tests/test_buildozer.py::TestBuildozer::test_buildozer_base PASSED                                                                                                                [ 14%]
tests/test_buildozer.py::TestBuildozer::test_buildozer_help PASSED                                                                                                                [ 21%]
tests/test_buildozer.py::TestBuildozer::test_buildozer_read_spec PASSED                                                                                                           [ 28%]
tests/test_buildozer.py::TestBuildozer::test_log_get_set PASSED                                                                                                                   [ 35%]
tests/test_buildozer.py::TestBuildozer::test_p4a_recommended_android_ndk_error PASSED                                                                                             [ 42%]
tests/test_buildozer.py::TestBuildozer::test_p4a_recommended_android_ndk_found PASSED                                                                                             [ 50%]
tests/test_buildozer.py::TestBuildozer::test_p4a_recommended_ndk_version_default_value PASSED                                                                                     [ 57%]
tests/test_buildozer.py::TestBuildozer::test_run_command_unknown PASSED                                                                                                           [ 64%]
tests/test_buildozer.py::TestCopyApplicationSources::test_exclude_dirs_filter PASSED                                                                                              [ 71%]
tests/test_buildozer.py::TestCopyApplicationSources::test_exclude_patterns PASSED                                                                                                 [ 78%]
tests/test_buildozer.py::TestCopyApplicationSources::test_ignore_hidden_files_and_directories_in_source PASSED                                                                    [ 85%]
tests/test_buildozer.py::TestCopyApplicationSources::test_include_extensions_filter PASSED                                                                                        [ 92%]
tests/test_buildozer.py::TestCopyApplicationSources::test_source_dir_with_hidden_parent PASSED                                                                                    [100%]

================================================================================== 14 passed in 0.16s ===================================================================================
(buildozer-py311) buildozer [patch-1 ↑·1|✔] $
```

So I push this and let you check it.

tests/test_buildozer.py
```diff
@@ -179,3 +181,280 @@ def test_p4a_recommended_android_ndk_found(
...
+        os.makedirs(os.path.dirname(filepath), exist_ok=True)
+        with open(filepath, 'w') as f:
+            f.write(content)
+        return filepath
+
+    # def create_dir(self, relpath):
+    #     """Helper to create a directory in source_dir."""
+    #     dirpath = os.path.join(self.source_dir, relpath)
+    #     os.makedirs(dirpath, exist_ok=True)
+    #     return dirpath
```
AndreMiras: Let's remove this commented out code completely

AndreMiras (approved):
Thanks for taking a stab at it.
I didn't expect integration tests to be honest, but that also works I guess.
Let's maybe just remove the unused commented out code and I think we're good for merging.
Thanks again :bow:

PR #1986, opened 2025-11-19
Split default spec into sections

kuzeyron:
First attempt on splitting the default spec so the user will only be greeted with the minimal spec file if just running `buildozer init`.

Usage:
    - `buildozer init`
    - `buildozer init android`
    - `buildozer init android --no-comments`
    - `buildozer init ios`
    - `buildozer init ios --no-docs`
    - `buildozer init osx`

Each options includes the documentation and default text to the `buildozer.spec` file and also aware not to include duplicates just to make it easier for the user.

These steps involves making it easier in the future for the developers to change default values without messing up the documentations.

With this PR I have also updated the default values to match our daily usage with Buildozer.

### DIFF
--- a/buildozer/default-android.json
+++ b/buildozer/default-android.json
@@ -283,7 +283,7 @@
             "doc": [
                 "(str) Android NDK version to use"
             ],
-            "default": "25b"
+            "default": "28c"
         },
         "android.ndk_api": {
             "doc": [


kuzeyron:
PR is ready for review.

T-Dynamos (approved):
LGTM - Great work!

akshayaurora:
Hiding the defaults would add to confusion just for the purpose of making it simpler for those who don't need the options.

I would much rather retain the default as is.

The purpose of default spec is to ::

- Offer a default specification which contains all features.
 - User can edit the things they want ignoring other options they do not need to.
 
 This PR goes against those basic tenets /principals of what a "Specification" is supposed to do/be.
 
 I would suggest retaining the current behavior for the default spec which shows all options.
 For a simpler minimal spec one could  add something like ::
 
    >> buildozer init minimal
    >> buildozer init <- retains existing behavior.

setup.py
```diff
@@ -79,7 +78,7 @@ def find_version(*file_paths):
         # here.
         # Restricted version because python-for-android's recipes can't handle
         # later versions.
-        'cython<3.0'
+        'cython<3.12'
```
AndreMiras: Are we sure we want to pull some of these changes in this pull request?
Some may deserve its own pull request to discuss it and to make it easier to spot regression and potentially revert

kuzeyron:
> Thanks for picking that up @kuzeyron 🙏 and thanks to both @T-Dynamos and @akshayaurora for reviewing. I think @akshayaurora may have a good point in a way that we don't break the default behavior at least meanwhile still making it possible to address your use case of minimal spec file. The only thing is to change the option to make it explicit we want it minimal, but the default stays as it was. What do you think?

I chose to update them because I had to bump other versions as someone had only updated the default spec but never the default value in the code. Buildozer haven't been working properly and I have addressed some of that in this PR. If this PR gets the green light, the next PR will cover the rest so in the future the person who changes versions cannot mess things up.

Intention of the default spec is to read options that we can override and if the user leave options out, Buildozer defaults to whatever the code says in `__init__.py`. And that's the proper way of doing it. So in the end osx, ios and android is handled correctly with this PR.

PR #2010, opened 2026-03-21
Add new p4a opts

T-Dynamos:
Ref: https://github.com/kivy/python-for-android/pull/3280

### DIFF
--- a/buildozer/default.spec
+++ b/buildozer/default.spec
@@ -48,7 +48,7 @@ requirements = python3,kivy==2.3.1
 # These indexes are searched in addition to the default PyPI index during Android builds.
 #  WARNING: Third party indexes are untrusted sources and may introduce supply chain risks,
 # including malicious, tampered, outdated, or incompatible packages.
-extra_index_urls = https://chaquo.com/pypi-13.1/, https://anshdadwal.is-a.dev/p4a-wheels/p4a/
+extra_index_urls = https://chaquo.com/pypi-13.1/
 
 # (bool) Disable use of prebuilt binary packages when available (currently Android only)
 # If enabled, all dependencies are built from source even if prebuilt wheels exist.
--- a/buildozer/default.spec
+++ b/buildozer/default.spec
@@ -42,7 +42,24 @@ version = 0.1
 
 # (list) Application requirements
 # comma separated e.g. requirements = sqlite3,kivy
-requirements = python3,kivy
+requirements = python3,kivy==2.3.1
+
+# (list) Additional package index urls used for dependency resolution (currently Android only)
+# These indexes are searched in addition to the default PyPI index during Android builds.
+#  WARNING: Third party indexes are untrusted sources and may introduce supply chain risks,
+# including malicious, tampered, outdated, or incompatible packages.
+extra_index_urls = https://chaquo.com/pypi-13.1/, https://anshdadwal.is-a.dev/p4a-wheels/p4a/
+
+# (bool) Disable use of prebuilt binary packages when available (currently Android only)
+# If enabled, all dependencies are built from source even if prebuilt wheels exist.
+skip_prebuilt = False
+
+# (list) Packages allowed to use available prebuilt wheels (currently Android only)
+# Overrides recipe pinned versions and bypasses default version selection behavior
+# when a compatible prebuilt wheel is available in the index.
+# Example: if kivy==2.3.1 is requested but only `3.0.0` is available in prebuilt index,
+# this option allows using the prebuilt wheel instead of building from source.
+# use_prebuilt_version_for = numpy,kivy
 
 # (str) Custom source folders for requirements
 # Sets custom source for any requirements with recipes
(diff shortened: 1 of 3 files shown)


buildozer/targets/android.py
```diff
@@ -123,6 +123,22 @@ def __init__(self, *args, **kwargs):
         if self.logger.log_level >= 2:
             self.extra_p4a_args.append("--debug")
 
+        save_wheel_dir = self.buildozer.config.getdefault('app', "p4a.save_wheel_dir", "")
+        if save_wheel_dir != "":
```
kuzeyron, line 127: `!= ""` is not needed. Just use `if save_wheel_dir:`

kuzeyron (commented):
Just one suggestion.

buildozer/default.spec
```diff
@@ -42,7 +42,24 @@ version = 0.1
...
 # (list) Application requirements
 # comma separated e.g. requirements = sqlite3,kivy
-requirements = python3,kivy
+requirements = python3,kivy==2.3.1
+
+# (list) Additional package index urls used for dependency resolution (currently Android only)
+# These indexes are searched in addition to the default PyPI index during Android builds.
+#  WARNING: Third party indexes are untrusted sources and may introduce supply chain risks,
+# including malicious, tampered, outdated, or incompatible packages.
+extra_index_urls = https://chaquo.com/pypi-13.1/, https://anshdadwal.is-a.dev/p4a-wheels/p4a/
```
kuzeyron, line 51: Are these trusted?

T-Dynamos: I think it's not my decision, I think it should be empty by default - until kivy hosts it's own. 
The index that I am hosting isn't granted to be stable (or secure?). It currently hosts every `PyProjectRecipe` and makes builds faster for me.

buildozer/default.spec
```diff
@@ -42,7 +42,24 @@ version = 0.1
...
+
+# (list) Additional package index urls used for dependency resolution (currently Android only)
+# These indexes are searched in addition to the default PyPI index during Android builds.
+#  WARNING: Third party indexes are untrusted sources and may introduce supply chain risks,
+# including malicious, tampered, outdated, or incompatible packages.
+extra_index_urls = https://chaquo.com/pypi-13.1/
+
+# (bool) Disable use of prebuilt binary packages when available (currently Android only)
+# If enabled, all dependencies are built from source even if prebuilt wheels exist.
+skip_prebuilt = False
```
AndreMiras, line 55: Shouldn't we have the default match the current buildozer behavior which is to build from source and not use prebuilt binaries?
And actually I have the same question for the (now merged) p4a pull request (https://github.com/kivy/python-for-android/pull/3280)

buildozer/default.spec
```diff
@@ -42,7 +42,24 @@ version = 0.1
...
 # (list) Application requirements
 # comma separated e.g. requirements = sqlite3,kivy
-requirements = python3,kivy
+requirements = python3,kivy==2.3.1
+
+# (list) Additional package index urls used for dependency resolution (currently Android only)
+# These indexes are searched in addition to the default PyPI index during Android builds.
+#  WARNING: Third party indexes are untrusted sources and may introduce supply chain risks,
+# including malicious, tampered, outdated, or incompatible packages.
+extra_index_urls = https://chaquo.com/pypi-13.1/
```
AndreMiras, line 51: Don't we miss the `p4a.*` namespace here and in the other options?
So `p4a.extra_index_urls` here for instance

T-Dynamos: This was done keeping in mind that this option in future will also be implemented for iOS.

AndreMiras, line 51: Also agree with @kuzeyron that maybe we shouldn't enable/ship an "untrusted" thirdparty index by default. Maybe keeping it commented out would be good enough

buildozer/targets/android.py
```diff
@@ -123,6 +123,22 @@ def __init__(self, *args, **kwargs):
...
+            self.extra_p4a_args.append(f"--save-wheel-dir={save_wheel_dir}")
+
+        extra_index_urls = self.buildozer.config.getlist('app', "extra_index_urls", "")
+        for index in extra_index_urls:
+            self.extra_p4a_args.append(f"--extra-index-url={index}")
+
+        use_prebuilt_version_for = self.buildozer.config.getlist('app', "use_prebuilt_version_for", "")
+        for _recipes in use_prebuilt_version_for:
```
AndreMiras, line 135: nit: I'm not sure we need the leading underscore, I would drop it

buildozer/targets/android.py
```diff
@@ -123,6 +123,22 @@ def __init__(self, *args, **kwargs):
         if self.logger.log_level >= 2:
             self.extra_p4a_args.append("--debug")
 
+        save_wheel_dir = self.buildozer.config.getdefault('app', "p4a.save_wheel_dir", "")
+        if save_wheel_dir != "":
+            self.extra_p4a_args.append(f"--save-wheel-dir={save_wheel_dir}")
+
+        extra_index_urls = self.buildozer.config.getlist('app', "extra_index_urls", "")
```
AndreMiras, line 130: Why don't we default to `[]` for both `getlist()` calls?

PR #2020, opened 2026-05-05
Update installation.rst for Ubuntu 26.04

RobertFlatt:
Simpler than 24.04 instructions as the default Ubuntu Python is now 3.14.


AndreMiras (approved):
LGTM thanks

PR #2024, opened 2026-05-09
Docs: Pin `sphinxawesome_theme` to 5.3.2

Novfensec:
Pinning sphinxawesome_theme version because of this!!
image

After pinning it works normally: (https://buildozer-fork.readthedocs.io/en/latest/installation/)
image

### DIFF
--- a/.github/workflows/test_python.yml
+++ b/.github/workflows/test_python.yml
@@ -68,7 +68,7 @@ jobs:
     steps:
     - uses: actions/checkout@v6
     - name: Requirements
-      run: pip install --upgrade sphinx sphinxawesome_theme
+      run: pip install --upgrade sphinx sphinxawesome_theme==5.3.2
     - name: Check links
       run: sphinx-build -b linkcheck docs/source docs/build
     - name: Generate documentation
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -83,7 +83,7 @@ def find_version(*file_paths):
     ],
     extras_require={
         'test': ['pytest'],
-        'docs': ['sphinx', 'sphinxawesome_theme'],
+        'docs': ['sphinx', 'sphinxawesome_theme==5.3.2'],
```
AndreMiras: Could we pin for major version instead e.g. `>=5.3,<6`

.github/workflows/test_python.yml
```diff
@@ -68,7 +68,7 @@ jobs:
     steps:
     - uses: actions/checkout@v6
     - name: Requirements
-      run: pip install --upgrade sphinx sphinxawesome_theme
+      run: pip install --upgrade sphinx sphinxawesome_theme==5.3.2
```
AndreMiras: Let's use this as an opportunity to DRY maybe with:
```
pip install --upgrade ".[docs]"
```

AndreMiras (commented):
Thanks for flagging this one! I've made a few suggestions, let me know what you think

AndreMiras (approved):
looking good thanks

PR #2039, opened 2026-05-24
p4a command update

WHG555:
When you run this command

```
buildozer android p4a clean_recipe_build pillow
```
This function takes as input ['p4a', 'clean_recipe_build', 'pillow']
```
def run_commands(self, args):
```

but can't run success.

### DIFF
--- a/buildozer/target.py
+++ b/buildozer/target.py
@@ -69,6 +69,10 @@ class Target:
                     result.append(last_command)
                     last_command = []
                 last_command.append(arg)
+                if arg == "p4a":
+                    while args:
+                        tmp = args.pop(0)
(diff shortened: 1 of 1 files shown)


AndreMiras:
Thanks for the contribution.
This is covered with that syntax `buildozer android p4a -- clean_recipe_build pillow`, note the `--` forwarding syntax.
Is this PR intended to add support for omitting the -- separator, or were you unaware of the existing syntax? If separator-free invocation is the intended enhancement, it needs to preserve the documented `--` form and include tests for both forms.
I'm adding a test to make sure we won't introduce that regression in the future https://github.com/kivy/buildozer/pull/2045

PR #2052, opened 2026-08-24
Provide support for building in a rootless container with Podman

ydirson:
Docker by default runs as root (relying on a system service so it may not be obvious to users), and your container can write files owned by the host's root user, which violates the principle of least priviledge; it can be made to run rootless but that's not my battle.

Podman OTOH does the right thing by running rootless.  By default however it maps the host user account running it to root in the container, and Buildozer does not like running as root, so we have to tell it to map the container non-priviledged user to the user account on the host (--userns=keep-id).

This allows the entrypoint script to be as bare-bones as possible, the Dockerfile achieves complete preparation for this.

### DIFF
--- a/podman-rootless/Dockerfile
+++ b/podman-rootless/Dockerfile
@@ -1,0 +1,99 @@
+# Dockerfile for providing buildozer in a rootless container using Podman
+#
+# Build with:
+#  podman build -f podman-rootless/Dockerfile \
+#    --tag=docker.io/kivy/buildozer:rootless \
+#    .
+#
+# In order to give the container access to your current working
+# directory it must be mounted using the --volume option.  To have the
+# unprivileged user mapped to your account so no shennanigans are
+# needed, we must use --userns=keep-id.
+#
+# The shell in the container is started as "user" by default, it you want to run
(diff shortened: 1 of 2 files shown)


ydirson (commented):
It could make sense to use a single `container` directory to host both rootfull-docker and rootless-podman material, wdyt?

ydirson:
Repushed without changes to original docker files that should have been thrown away

podman-rootless/Dockerfile
```diff
@@ -0,0 +1,99 @@
...
+# directory it must be mounted using the --volume option.  To have the
+# unprivileged user mapped to your account so no shennanigans are
+# needed, we must use --userns=keep-id.
+#
+# The shell in the container is started as "user" by default, it you want to run
+# something as root ou can start it with --user=root (and use "su user" if needed).
+#
+# Run with (e.g. `buildozer --version`):
+# podman run --interactive --tty --rm \
+#   --userns=keep-id \
```
ydirson: There is an assumption here that the user's uid is the one of the in-container user.

ydirson (commented):
Setting this PR to Draft until I get this fixed, should not be that hard to play with the uidmap.

